PR #1, opened 2026-03-05
fix: handle io.ReadAll errors and add findings store capacity limits

haasonsaas:
## Summary

- **Fix silent error suppression**: Six instances of `body, _ := io.ReadAll(resp.Body)` across Anthropic, OpenAI, and GitLab clients now handle read errors instead of silently producing empty error messages (e.g., `API error 429: ` with no body detail)
- **Add findings store capacity limits**: The in-memory `Store` now supports `MaxFindings` via `NewStoreWithConfig()`, evicts oldest resolved findings when capacity is exceeded, and exposes `Cleanup(maxAge)` and `Len()` methods
- **Backward compatible**: `NewStore()` remains unlimited; all existing tests pass alongside 4 new tests

## Test plan

- [x] `make build` compiles cleanly
- [x] `go test ./...` — all existing + 4 new tests pass
- [x] `go vet ./...` — no issues
- [ ] Verify error messages in staging with a simulated 429/500 response
- [ ] Validate store eviction under load in a long-running instance

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -261,14 +261,17 @@ jobs:
         run: docker build -f Dockerfile -t cerebro:ci .
 
       - name: Run Trivy image scan
-        uses: aquasecurity/trivy-action@0.34.0
-        with:
-          image-ref: cerebro:ci
-          format: table
-          exit-code: '1'
-          ignore-unfixed: true
-          vuln-type: os,library
-          severity: CRITICAL,HIGH
+        run: |
+          docker run --rm \
+            -v /var/run/docker.sock:/var/run/docker.sock \
+            -v "${HOME}/.cache/trivy:/root/.cache/trivy" \
+            aquasec/trivy:0.69.1 image \
+            --format table \
+            --exit-code 1 \
+            --ignore-unfixed \
+            --vuln-type os,library \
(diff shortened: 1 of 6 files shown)


jonathanhaaswriter (changes requested):
Blocking issue:  is described as a capacity limit, but the current logic only evicts RESOLVED findings. If all findings are OPEN, the store exceeds the configured max (repro: max=2, upsert 3 open findings -> len=3). This means memory can still grow unbounded in exactly the high-churn scenarios this change targets.

Please either (a) enforce a true hard cap with a fallback eviction policy when no resolved findings are available, or (b) rename/scope the setting to make it explicit that only resolved findings are bounded.

Non-blocking:  is introduced but not used yet, which may be confusing to callers.

jonathanhaaswriter (approved):
Addressed.

- `StoreConfig.MaxFindings` now enforces a true hard cap even when all findings are OPEN.
- Eviction order is deterministic and status-aware: RESOLVED, then SUPPRESSED, then other statuses, oldest `LastSeen` first.
- `StoreConfig.ResolvedRetention` is now active and applied during upserts.
- Added tests for hard-cap behavior and retention cleanup.

PR #12, opened 2026-03-06
Fix /metrics 500 from duplicate Prometheus collector registration

haasonsaas:
## Summary
- fix a runtime panic triggered when `/metrics` registers core collectors that collide with API-local `promauto` collectors
- rename API-local metric names to avoid global registry fqName collisions
- add regression test hitting `/metrics` to ensure endpoint stays healthy

## Reproduction before fix
- start API with auth enabled
- call `/metrics` with a valid API key
- observed `500` and panic in logs: duplicate descriptor for `cerebro_scan_duration_seconds`

## Validation
- go test ./internal/api/...
- manual runtime check: authenticated `/metrics` now returns 200

### DIFF
--- a/internal/api/metrics.go
+++ b/internal/api/metrics.go
@@ -15,7 +15,7 @@ import (
 var (
 	httpRequestsTotal = promauto.NewCounterVec(
 		prometheus.CounterOpts{
-			Name: "cerebro_http_requests_total",
+			Name: "cerebro_api_http_requests_total",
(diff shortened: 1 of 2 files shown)


haasonsaas:
Stack extension note:
- #13 and #14 are based on this branch (fix/metrics-endpoint-duplicate-registry-panic).
- Merge #12 first, then merge #13/#14 in any order.

PR #13, opened 2026-03-06
Auth/OpenAPI parity: keep docs, spec, and metrics public

haasonsaas:
## Summary
- align runtime auth behavior with the OpenAPI contract for public endpoints
- treat `/docs`, `/openapi.yaml`, and `/metrics` as public endpoints in API key auth middleware
- deduplicate public-endpoint checks between API auth and RBAC middleware
- extend API auth smoke CI to assert those public endpoints stay accessible without auth
- add API tests for public endpoint behavior when auth is enabled

## Why
OpenAPI currently marks these routes with `security: []`, but runtime auth previously returned 401 for them when auth was enabled.

## Validation
- go test ./internal/api/...
- make openapi-check
- parsed .github/workflows/ci.yml with Python YAML loader
- manual smoke: unauthenticated `/metrics`, `/docs`, `/openapi.yaml` return 200 while protected API routes remain auth-gated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -308,6 +308,16 @@ jobs:
             exit 1
           fi
 
+          for public_path in /metrics /docs /openapi.yaml; do
+            public_status=$(curl -sS -o "/tmp/api-auth-smoke-public-$(echo "$public_path" | tr '/.' '__').json" -w "%{http_code}" \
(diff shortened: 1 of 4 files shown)


haasonsaas:
Stack dependency note:
- Base branch is #12 (fix/metrics-endpoint-duplicate-registry-panic).
- Merge #12 first.

PR #16, opened 2026-03-06
api: align rate limiting with public endpoint and auth key parsing

haasonsaas:
## Summary
- use the shared `isPublicEndpoint` helper in rate-limit middleware to avoid drift from auth/RBAC public endpoint rules
- canonicalize rate-limit API-key extraction via `extractAPIKeyStrict` so `Authorization: Bearer <key>` and `X-API-Key: <key>` share one bucket
- expand rate-limit tests to cover `/docs` and `/openapi.yaml` bypass behavior plus malformed/conflicting credential fallback to IP keying

## Testing
- go test ./internal/api/...
- go test ./...


haasonsaas:
Stack update: follow-up PR is up at #17 (base: this branch) to normalize rate-limit client IP extraction and prevent source-port-based bypass.

PR #18, opened 2026-03-06
test: load openapi contract spec from embedded bytes

haasonsaas:
## Summary
- switch `internal/api/openapi_contract_test.go` to load the OpenAPI spec from embedded bytes (`apicontract.OpenAPIYAML`) instead of relative file reads
- make `TestNoPlaceholderMarkersInAPIServer` resolve `server.go` via `runtime.Caller` + `filepath.Join` so it does not depend on process working directory
- improve API test resilience by removing cwd-sensitive path assumptions

## Testing
- go test ./internal/api/...
- go test ./...

### DIFF
--- a/internal/api/server_test.go
+++ b/internal/api/server_test.go
@@ -10,6 +10,8 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"path/filepath"
+	goruntime "runtime"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


haasonsaas:
Stack update: follow-up PR is up at #19 (base: this branch) to harden  query parsing and prevent non-positive values from flowing into audit/data handlers.

PR #28, opened 2026-03-10
feat: Add new GCP/AWS IAM permission usage scans

josh-writer:


jonathanhaaswriter (changes requested):
Thanks for pushing this. I found a few blocking issues before merge:

1. `internal/sync/tables_identitycenter_usage.go:350-377, 585-605` — policy documents from `GetPolicyVersion` are URL-encoded, but this path sends the raw document to `graph.ParseAWSPolicy`. Existing code in `internal/sync/tables_iam_policies.go:112-115` unescapes first. As written, attached policies can resolve to zero actions and make permission sets look empty or unused.
2. `internal/sync/tables_identitycenter_usage.go:676-735` — `GetServiceLastAccessedDetails` is paginated via `IsTruncated`/`Marker`, but the implementation only reads the first completed page. Large roles will miss later services/actions and can be mislabeled as unused.
3. `internal/sync/tables_gcp_iam_group_usage.go:51-55` + `internal/sync/gcp.go:361` + `internal/sync/scoped_table_ops.go:28-35` — the table is always registered, but when `--gcp-iam-target-groups` is unset the fetcher returns `nil`; the scoped upsert treats that as authoritative empty data and deletes existing rows for the project.
4. `internal/sync/tables_identitycenter_usage.go:111-120` + `internal/sync/engine.go:493-496` — this table is incremental, so only current row IDs are replaced. Permission sets skipped by include/exclude filters, or deleted from Identity Center entirely, never get cleaned up and stale rows remain.
5. `internal/sync/tables_gcp_iam_group_usage.go:137-180, 322-328` — the GCP logic marks a group permission as used if any member used that permission in audit logs, but it does not prove the permission came from that group’s binding versus another grant or a direct role. That can hide unused group grants and make the recommendation misleading.

jonathanhaaswriter (changes requested):
Thanks for the PR — the feature direction looks good, but I found three blocking correctness issues that should be fixed before merge:

1. AWS Identity Center is hard-coded to `us-east-1`.
   - `internal/sync/aws_limits.go:10-17` forces all `aws_identitycenter_*` tables to `us-east-1`.
   - `internal/sync/tables_identitycenter_usage.go:66-83` then calls `ListInstances` in that region and silently skips on failure/no instances.
   Identity Center is regional and many tenants use a different home region, so this will silently miss all permission-usage data outside `us-east-1`.
   Suggested fix: detect/probe the Identity Center home region instead of forcing `us-east-1`.

2. Transient IAM policy lookup failures can delete valid AWS permission-usage rows.
   - `internal/sync/tables_identitycenter_usage.go:340-379` swallows `GetPolicy` / `GetPolicyVersion` failures with `continue`.
   - The caller still treats `grantedActions` as authoritative and runs stale-row deletion in `internal/sync/tables_identitycenter_usage.go:151-166` and `:259`.
   If one attached policy lookup fails, the code can drop actions from that policy or even wipe all rows for the permission set.
   Suggested fix: surface incomplete policy resolution and skip stale-row deletion whenever the granted-action set is not fully resolved.

3. GCP empty-result paths are treated as authoritative and purge project-scoped data.
   - `internal/sync/tables_gcp_iam_group_usage.go:51-69` returns no rows when no target groups are configured.
   - `internal/sync/tables_gcp_iam_group_usage.go:243-295` also creates empty target groups that later get skipped.
   - `internal/sync/gcp.go:293-295` scopes the upsert to the project, and `internal/sync/scoped_table_ops.go:29-40` deletes all scoped rows when `rows` is empty.
   That means missing config, a target group with no current binding, or similar non-authoritative states can erase previously valid data for the whole project.
   Suggested fix: distinguish authoritative-empty results from skipped/incomplete/not-configured states and avoid scoped deletion in the latter cases.

jonathanhaaswriter (approved):
Latest pass looks good to me. Approving pending final CI completion.

PR #59, opened 2026-03-16
fix(ci): stabilize time-sensitive main checks

jonathanhaaswriter:
## Summary
- replace date-sensitive `last=7d` changelog test windows with explicit `since`/`until` ranges
- make the freshness confidence test relative to the current clock so it stays meaningful over time
- pin and shard the scheduled source gosec scan to avoid the flaky runner-shutdown behavior on main

## Validation
- `go test ./... -count=1`
- `$(go env GOPATH)/bin/golangci-lint run --timeout 15m ./cmd/... ./internal/... ./api/...`

### DIFF
--- a/.github/workflows/security-source-scan.yml
+++ b/.github/workflows/security-source-scan.yml
@@ -10,8 +10,14 @@ permissions:
 
 jobs:
   gosec-source-scan:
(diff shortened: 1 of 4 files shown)


prestrike00 (approved):
LGTM

PR #147, opened 2026-03-24
feat(graph): prefer scoped backend reads

jonathanhaaswriter:
## Summary
- add a tenant-scoped read-only graph-store wrapper and prefer configured tenant-scoped backends before snapshot-backed stores
- route blast-radius plus non-temporal graph-query neighbors and paths through `GraphStore` for both the tool surface and the platform API
- track the Neptune-native runtime refactor status in `TODO.md` and cover the new store-native paths with tests

## Validation
- go test ./internal/app ./internal/api ./internal/graph
- python3 scripts/devex.py run --files TODO.md internal/api/server_handlers_graph_intelligence.go internal/api/server_handlers_graph_intelligence_test.go internal/api/server_services_graph_intelligence.go internal/app/app_cerebro_tools.go internal/app/app_cerebro_tools_test.go internal/app/app_graph_store.go internal/app/app_graph_store_test.go internal/app/app_graph_view.go internal/graph/store_dual_write.go internal/graph/store_neptune.go internal/graph/store_tenant_scoped.go internal/graph/store_tenant_scoped_test.go
- python3 scripts/oss_audit.py

## Context
This is the first foundation slice toward running Neptune as the primary graph store without hidden full-graph materialization on supported read paths.


jonathanhaaswriter:
I think we still have two issues here:

- the synthetic snapshot-record fallback hashes `BuiltAt`, but `graphMetadataFromCounts()` uses `time.Now()` when only counts are available, so `/platform/graph/snapshots/current` can generate a different snapshot ID on every identical read
- the new configured tenant-aware backend path appears to bypass the existing missing-tenant guard, so it can return shared nodes for nonexistent tenants instead of failing closed

Can we avoid generating unstable snapshot IDs from `time.Now()`, and apply the same tenant-presence check to configured tenant-aware backends?

jonathanhaaswriter:

PR #149, opened 2026-03-24
fix(app): allow nlq previews without graph

jonathanhaaswriter:
## Summary
- allow `cerebro.nl_query` preview-only requests to translate questions without requiring a readable security graph
- keep execution mode unchanged so non-preview NLQ requests still fail fast when the graph is unavailable
- add TDD coverage for the preview-only contract

## Validation
- go test ./internal/app
- python3 scripts/devex.py run --files internal/app/app_cerebro_tools_nlq.go internal/app/app_cerebro_tools_test.go

## Stack
Base branch: `feat/graph-store-read-foundation`


jonathanhaaswriter:
I think the tenant fail-closed contract regressed here. Once the configured Spanner store advertises tenant read scope, `CurrentSecurityGraphStoreForTenant()` prefers that path without the `HasScopedNodesForTenant()` guard the other tenant resolvers use. Because shared nodes are still visible under tenant scope, a `tenant-missing` request can return shared data instead of `ErrStoreUnavailable`.

Can we add the same tenant-presence check (or a Spanner equivalent) before returning the scoped store?

PR #151, opened 2026-03-24
fix(api): prefer live graphs for visualizations

jonathanhaaswriter:
## Summary
- use the current tenant graph view for attack-path, toxic-combination, and report Mermaid endpoints
- avoid unnecessary snapshot reads when a live graph is already available
- add regressions to prove these visualizations stay off the snapshot store in live-graph mode

## Testing
- go test ./internal/api -count=1


jonathanhaaswriter:

PR #152, opened 2026-03-24
fix(api): read org policy views from subgraphs

jonathanhaaswriter:
## Summary
- route per-policy org-policy reads through bounded subgraphs instead of full tenant graph materialization
- prefer the live tenant graph when available and fall back to store-native ExtractSubgraph reads otherwise
- add regressions for both store-only and live-graph execution paths

## Testing
- go test ./internal/api -count=1


jonathanhaaswriter:

PR #155, opened 2026-03-25
feat(api): use store subgraphs for platform knowledge detail reads

jonathanhaaswriter:
## Summary
- route platform knowledge detail reads through bounded store subgraphs instead of whole-graph materialization
- cover claim, evidence, observation, explanation, timeline, and proofs handlers with store-backed/live-graph tests
- keep existing error behavior when the platform graph is unavailable

## Testing
- go test -count=1 ./internal/api
- /Users/jonathan/go/bin/golangci-lint run --timeout 5m ./internal/api
- make openapi-check


jonathanhaaswriter:
I think the fixed `MaxDepth: 3` subgraph in `claimGraph(...)` is too small for the recursive detail endpoints. `GetClaimTimeline`, `ExplainClaim`, and `BuildClaimProofs` all follow related claims transitively, so a chain like `c4 -> c3 -> c2 -> c1 -> root` gets silently truncated once the support/refute/proof graph extends beyond 3 hops.

Could we avoid the fixed-depth subgraph for these recursive detail reads, or size it based on the requested expansion depth?

jonathanhaaswriter:

PR #159, opened 2026-03-25
Migrate storage layer from Snowflake/DynamoDB/SQS to Postgres/NATS (graph backend fixes)

jonathanhaaswriter:
## Summary
- start from `feat/postgres-nats-migration` and keep Postgres scoped to storage and warehouse responsibilities instead of treating it as a graph backend
- move read-only SQL validation into backend-agnostic warehouse helpers and update API, agent, and CLI call sites accordingly
- wire in the pgx stdlib driver plus vendored dependencies required for Postgres runtime support, and refresh generated config env docs

## Testing
- `CEREBRO_DEVEX_BASE_REF=writer/main python3 ./scripts/devex.py run --mode changed --base-ref writer/main`


jonathanhaaswriter:
I think we still have two problems here:

- `effectivePostgresURL()` prefers `JOB_DATABASE_URL` over `DATABASE_URL`, so if both are set the app repos silently connect to the jobs DB instead of the normal app DB
- the local findings fallback changed from SQLite to in-memory, which drops the persisted local-mode behavior we had before

Can we keep app storage on `DATABASE_URL` and preserve the SQLite fallback when Postgres isn't configured?

PR #160, opened 2026-03-25
Refactor SecurityTools to use Warehouse interface instead of Snowflak…

josh-writer:
…e. Update related tests and application initialization to reflect this change. Enhance error messages for asset queries to specify warehouse configuration requirements. Additionally, import pgx driver in main.go for PostgreSQL support. Replace sqs with NATS.


jonathanhaaswriter (changes requested):
I found three blocking issues:

1. `cmd/cerebro/main.go` now imports `github.com/jackc/pgx/v5/stdlib`, but `internal/app/app_init_core.go` still calls `sql.Open("postgres", databaseURL)`. The pgx stdlib driver registers `pgx` / `pgx/v5`, not `postgres`, so this still fails at runtime with `sql: unknown driver "postgres"`.

2. The vendored AWS removals leave the repo in an inconsistent vendor state. This PR deletes `vendor/github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue/*` and `vendor/.../service/dynamodbstreams/*`, but does not update `go.mod`, `go.sum`, or `vendor/modules.txt` to match. In this repo that breaks `make vendor-check` / `-mod=vendor` builds.

3. `internal/app/app_security_services.go` removes the second `requireGraphWriterLease("rebuild security graph")` check in `RebuildSecurityGraph` right before activating the rebuilt graph. That re-check protects against losing the writer lease during a long rebuild; without it, a stale builder can still publish over a newer writer.

jonathanhaaswriter (commented):
All three issues addressed in `f1d1bad31`:

1. **Driver name** — Changed `sql.Open("postgres", databaseURL)` to `sql.Open("pgx", databaseURL)` in `app_init_core.go`. The `pgx/v5/stdlib` import is now co-located with usage in `internal/app/pgx_driver.go` and `internal/cli/pgx_driver.go`.

2. **Vendor state** — Removed `feature/dynamodb/attributevalue` and `service/dynamodbstreams` entries from `go.mod`, `go.sum`, and `vendor/modules.txt` to match the deleted vendor files. `go build` and `go vet` both pass clean with `-mod=vendor`.

3. **Writer lease check** — Restored the second `requireGraphWriterLease("rebuild security graph")` call at line 739, immediately before `activateBuiltSecurityGraph`. This re-validates the lease after the potentially long rebuild to prevent stale builders from publishing over a newer writer.

PR #176, opened 2026-03-27
fix(graph): remove local snapshot read fallbacks

jonathanhaaswriter:
## Summary
- remove the remaining local snapshot-backed graph read fallbacks from app graph resolution and runtime adapters
- make incremental graph updates fail fast instead of silently rebuilding from local snapshot state
- update graph/readiness tests to assert configured-store-only behavior for the strict Neptune cutover

## Validation
- make verify


jonathanhaaswriter:
I think `WaitForReadableSecurityGraph()` can still hand callers the empty live builder graph. The new `CurrentSecurityGraph()` logic correctly falls back to the configured graph when the live graph is empty, but the wait path still returns `a.currentLiveSecurityGraph()` after `WaitForGraph()` succeeds. On lease followers / configured-store readers that means the wait succeeds because the store is readable, but callers can still get the empty in-memory graph back.

Can we return `a.CurrentSecurityGraph()` (or the stored view) after the wait instead of the raw live graph?

internal/app/app_graph_status.go
```diff
@@ -56,22 +56,17 @@ func (a *App) setGraphBuildState(state GraphBuildState, builtAt time.Time, err e
...
 	}
 	if a == nil {
 		return nil
 	}
 	if view, err := a.currentConfiguredSecurityGraphView(context.Background()); err == nil && view != nil {
 		return view
 	}
-	view, err := a.storedSecurityGraphViewWithSnapshotLoader(func(store *graph.GraphPersistenceStore) (*graph.Snapshot, error) {
-		snapshot, _, _, err := store.PeekLatestSnapshot()
-		return snapshot, err
```
josh-writer: After removing snapshot-backed reads, CurrentSecurityGraph() now falls back to a.currentLiveSecurityGraph() whenever currentConfiguredSecurityGraphView() returns an error or no view. On read-only replicas, initSecurityGraph() still publishes the builder’s empty graph before checking the writer lease (internal/app/app_security_services.go:528-536), so a Neptune outage/unready store now gets surfaced as a zero-node graph instead of ErrStoreUnavailable.

This is user-visible because API read paths prefer CurrentSecurityGraph() over the store (internal/api/graph_view_resolvers.go:9-12), so graph/risk/intelligence endpoints can silently return empty results during Neptune failures.

jonathanhaaswriter: Addressed in `f60ba0bb9`. `CurrentSecurityGraph` and `currentOrStoredSecurityGraphView` now return nil when the live graph is unreadable and the configured view is missing or errors, so API read paths fall back to the store and surface unavailability instead of a zero-node graph. I also added regression coverage for the missing-view and error paths and reran `python3 ./scripts/devex.py run --mode changed --base-ref writer/main`.

internal/app/app_graph_backend.go
```diff
@@ -58,5 +58,7 @@ func (a *App) probeConfiguredSecurityGraphStore(ctx context.Context, store graph
 	if err != nil {
 		return false, fmt.Errorf("probe configured graph store edges: %w", err)
 	}
-	return nodes > 0 || edges > 0, nil
+	_ = nodes
+	_ = edges
+	return true, nil
```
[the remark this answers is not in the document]
jonathanhaaswriter: The current branch still keeps zero-node configured backends unready: `probeConfiguredSecurityGraphStore` returns `nodes > 0 || edges > 0`, and `TestInitConfiguredSecurityGraphStoreLeavesEmptyBackendUnready` covers that guard. This thread is outdated relative to the current head, so I am resolving it.

PR #180, opened 2026-03-27
fix(scan): use warehouse-backed direct scans

jonathanhaaswriter:
## Summary
- switch direct scan and scan preflight from Snowflake-only wiring to the configured warehouse so sqlite/postgres backends can run scans without a Snowflake client
- add shared warehouse asset-query helpers that preserve `_cq_table`, dedupe on latest `_cq_sync_time`, and honor incremental/keyset scan cursors on sqlite/postgres
- add regression coverage for warehouse-backed direct scans plus sqlite incremental pagination/filter behavior

## Validation
- go test ./...
- GOTOOLCHAIN=go1.26.1 go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.8.0 run --timeout 5m ./...
- go run ./scripts/check_agent_sdk_contract_compat/main.go --require-baseline --base-ref=writer/main
- make agent-sdk-docs-check
- make agent-sdk-packages-check
- go run ./scripts/check_api_contract_compat/main.go --require-baseline --base-ref=writer/main
- make api-contract-docs-check
- make config-docs-check
- make devex-codegen-check
- go test ./internal/app -run "Test(PreCommitHookRunsFastLintOnStagedGoFiles|PrePushHookRunsChangedDevexPreflight|DockerBuildCommandsPassGoVersionBuildArg|DevelopmentGuideDocumentsDevexPreflight|DevexScriptPlansRelevantChecks|DevexScriptChangedModeIncludesWorkspaceDiffSources)"
- make openapi-check
- go run ./scripts/check_report_contract_compat/main.go --require-baseline --base-ref=writer/main
- make report-contract-docs-check
- make vendor-check

### DIFF
--- a/internal/app/app_scan_scheduler.go
+++ b/internal/app/app_scan_scheduler.go
@@ -271,7 +271,7 @@ func (a *App) runScheduledGraphAnalyses(ctx context.Context, tuning ScanTuning, 
 
 func (a *App) runScheduledScan(ctx context.Context, tables []string) error {
 	if a.Warehouse == nil {
-		return fmt.Errorf("snowflake not configured")
+		return fmt.Errorf("warehouse not configured")
 	}
 
 	scanStart := time.Now()
--- a/internal/cli/row_helpers.go
+++ b/internal/cli/row_helpers.go
@@ -1,0 +1,66 @@
+package cli
+
+import (
+	"strings"
+	"time"
+)
+
+func scheduleRowValue(row map[string]interface{}, key string) interface{} {
+	for candidate, value := range row {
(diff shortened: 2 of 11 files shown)


jonathanhaaswriter:
I think the new direct-scan path loses watermark progress for delete-only incremental batches. The current flow exits early when `GetAssets` returns no rows and only persists watermarks when `scanned > 0`, but delete-only CDC batches still need to advance the cursor even when there is nothing left to fetch. Otherwise the scan can keep replaying the same delete-only window.

Can we preserve the old "advance watermark from change events even when no assets are returned" behavior?

PR #181, opened 2026-03-27
fix(agents): use warehouse-backed query surfaces

jonathanhaaswriter:
## Summary
- switch agent query/get-asset tools and their CLI/job wiring from Snowflake clients to the generic warehouse read surface
- add a lightweight `app.OpenWarehouse` helper so direct `cerebro query` uses the configured warehouse backend without booting the full app
- add regression coverage for warehouse-backed agent queries and direct sqlite-backed CLI queries

## Validation
- go test ./...
- GOTOOLCHAIN=go1.26.1 go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.8.0 run --timeout 5m ./...
- go run ./scripts/check_agent_sdk_contract_compat/main.go --require-baseline --base-ref=writer/main
- make agent-sdk-docs-check
- make agent-sdk-packages-check
- go run ./scripts/check_api_contract_compat/main.go --require-baseline --base-ref=writer/main
- make api-contract-docs-check
- make config-docs-check
- make devex-codegen-check
- go test ./internal/app -run "Test(PreCommitHookRunsFastLintOnStagedGoFiles|PrePushHookRunsChangedDevexPreflight|DockerBuildCommandsPassGoVersionBuildArg|DevelopmentGuideDocumentsDevexPreflight|DevexScriptPlansRelevantChecks|DevexScriptChangedModeIncludesWorkspaceDiffSources)"
- make openapi-check
- go run ./scripts/check_report_contract_compat/main.go --require-baseline --base-ref=writer/main
- make report-contract-docs-check
- make vendor-check

### DIFF
--- a/internal/cli/query.go
+++ b/internal/cli/query.go
@@ -39,6 +39,21 @@ var (
 	runQueryDirectFn = runQueryDirect
 )
 
+var allowedReadOnlyPRAGMAs = map[string]struct{}{
+	"TABLE_INFO":       {},
+	"TABLE_XINFO":      {},
+	"TABLE_LIST":       {},
+	"INDEX_LIST":       {},
+	"INDEX_INFO":       {},
+	"INDEX_XINFO":      {},
+	"FOREIGN_KEY_LIST": {},
+	"DATABASE_LIST":    {},
+	"COMPILE_OPTIONS":  {},
+	"FUNCTION_LIST":    {},
+	"MODULE_LIST":      {},
+	"PRAGMA_LIST":      {},
+}
+
 func init() {
 	queryCmd.Flags().StringVarP(&queryFormat, "format", "f", "table", "Output format: table, json, csv")
 	queryCmd.Flags().IntVarP(&queryLimit, "limit", "l", 100, "Limit results")
@@ -90,9 +105,18 @@ func runQueryDirect(cmd *cobra.Command, args []string) error {
 	}
 
 	query := strings.Join(args, " ")
-	query, _, err = warehouse.BuildReadOnlyLimitedQuery(query, queryLimit)
-	if err != nil {
+	rawQuery := false
+	if metadataQuery, executeRaw, handled, err := prepareDirectMetadataQuery(query); err != nil {
 		return err
+	} else if handled {
+		query = metadataQuery
+		rawQuery = executeRaw
+	}
+	if !rawQuery {
+		query, _, err = warehouse.BuildReadOnlyLimitedQuery(query, queryLimit)
+		if err != nil {
+			return err
+		}
 	}
 
 	ctx, cancel := context.WithTimeout(commandContextOrBackground(cmd), warehouse.ClampReadOnlyQueryTimeout(int((60*time.Second)/time.Second)))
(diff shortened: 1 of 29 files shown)


internal/warehouse/query_validation.go
```diff
@@ -0,0 +1,272 @@
...
+				if nextIndex+1 < len(query) && query[nextIndex] == '(' {
+					close := nextIndex + 1
+					for close < len(query) && (query[close] == ' ' || query[close] == '\t' || query[close] == '\n' || query[close] == '\r') {
+						close++
+					}
+					if close < len(query) && query[close] == ')' {
+						after := close + 1
+						nextIsWord := after < len(query) && isWordChar(query[after])
+						if !nextIsWord {
+							out.WriteString("CURRENT_TIMESTAMP()")
```
josh-writer, line 276: Moving the read-only query builder into the generic warehouse path makes its partial dialect rewrite apply to SQLite/Postgres callers too, but it still only rewrites NOW() to CURRENT_TIMESTAMP(). Queries that use the existing policy pattern NOW() - INTERVAL 'N days' still go out as CURRENT_TIMESTAMP() - INTERVAL 'N days', which SQLite cannot execute. That will break the new direct SQLite query flow and warehouse-backed agent queries on common time-based predicates.

internal/providers/storage.go
josh-writer, line 1: This branch now routes provider sync through generic warehouses, but the storage path is still the variant/JSON path:  EnsureVariantTable creates JSONB/TEXT columns and MergeVariantRowsBatch JSON-marshals every provider field. On Postgres/SQLite that stores scalars as JSON values instead of native text/bool/number columns, so existing query/policy SQL like account_enabled = true, LOWER(email), or COALESCE(filevault_enabled, false) will fail or return no matches.

internal/cli/query.go
```diff
@@ -76,36 +78,27 @@ func runQuery(cmd *cobra.Command, args []string) error {
...
+		defer func() { _ = closer.Close() }()
 	}
-	defer func() { _ = client.Close() }()
 
 	query := strings.Join(args, " ")
-	upperQuery := strings.ToUpper(strings.TrimSpace(query))
-	// Only add LIMIT to SELECT queries that don't already have one
-	if strings.HasPrefix(upperQuery, "SELECT") && !strings.Contains(upperQuery, "LIMIT") && queryLimit > 0 {
-		query = fmt.Sprintf("%s LIMIT %d", query, queryLimit)
+	query, _, err = warehouse.BuildReadOnlyLimitedQuery(query, queryLimit)
```
josh-writer: runQueryDirect now forces every direct query through BuildReadOnlyLimitedQuery. That changes direct-mode semantics in two ways: explicit LIMIT clauses are silently clamped again by the default --limit=100, and safe metadata statements like SHOW TABLES, DESCRIBE TABLE, or SQLite PRAGMA table_info(...) are now rejected by the SELECT/WITH-only validator. The old direct path only appended a limit when one was missing and otherwise preserved the user’s read-only SQL.

josh-writer (changes requested):

PR #186, opened 2026-03-30
fix: complete Neptune and Postgres state cutover

jonathanhaaswriter:
## Summary
- remove local snapshot read fallbacks so graph reads stay on the live/configured runtime path
- migrate app-state persistence and Snowflake backfill paths to Postgres while preserving the legacy Snowflake migration source
- sync generated config env docs and stabilize time-windowed graph intelligence API tests

## Validation
- `env GOFLAGS=-mod=vendor go -C "/Users/jonathan/Documents/cerebro-worktrees/writer-main-neptune-pg-cutover-20260330" test ./internal/app ./internal/appstate ./internal/agents ./internal/findings ./internal/api ./internal/cli ./internal/snowflake ./internal/remediation ./scripts/generate_agent_sdk_packages -count=1`
- `python3 scripts/devex.py run --mode changed --base-ref writer/main`
- `python3 scripts/devex.py run --mode pr --base-ref writer/main`


josh-writer (changes requested):

internal/app/app_state_postgres.go
```diff
@@ -0,0 +1,489 @@
...
+			_ = db.Close()
+			return err
+		}
+	}
+
+	a.appStateDB = db
+	return nil
+}
+
+func (a *App) migrateAppState(ctx context.Context) error {
```
josh-writer, line 94: migrateAppState() returns early when appStateMigrationSnowflake() is nil, so pure Postgres cutovers skip migrateFindings() entirely. But migrateFindings() still needs to import legacy Postgres-backed findings before it ever consults Snowflake (internal/app/app_state_postgres.go:107-112).  

I think we should always run the Postgres findings migration when appStateDB is present; only gate the Snowflake-backed migrations individually.

internal/app/app_secrets.go
```diff
@@ -241,70 +241,74 @@ func (a *App) rotateSnowflakeClient(ctx context.Context, cfg *Config) error {
 		ctx = context.Background()
 	}
 
-	if strings.TrimSpace(cfg.SnowflakePrivateKey) == "" ||
-		strings.TrimSpace(cfg.SnowflakeAccount) == "" ||
-		strings.TrimSpace(cfg.SnowflakeUser) == "" {
+	if !hasSnowflakeCredentials(cfg) {
```
josh-writer, line 320: When Snowflake creds are removed on a non-Snowflake backend, rotateSnowflakeClient() now returns early without clearing any existing LegacySnowflake/Snowflake client. That stale client can still be used later, including by agent tool wiring (internal/app/app_init_agents.go:29-33). The new test only asserts config changes, not teardown (internal/app/app_secrets_test.go:145-185).  

We should explicitly close/clear legacy Snowflake state when creds are removed on a Postgres backend, and add a regression test that asserts the old client is no longer used.

jonathanhaaswriter:
Pushed `5ccb0acef` to finish the remaining follow-up work on this branch. In addition to the earlier cutover review fixes, this update swaps the Postgres `database/sql` driver from `pgx` stdlib to `lib/pq`, bumps the Go toolchain line to `go1.26.2`, refreshes vendored deps, and clears the new `govulncheck` blocker. I reran `python3 ./scripts/devex.py run --mode changed --base-ref writer/main` locally on the branch and it passed end-to-end; the fresh CI run is now in flight. @josh-writer could you take another pass when you have a moment?

josh-writer (approved):

PR #259, opened 2026-04-10
refactor(app): add subsystem lifecycle seams

jonathanhaaswriter:
## Summary
- add reusable subsystem lifecycle helpers for explicit init/start sequencing in `internal/app`
- introduce grouped subsystem-scoped config views for agents, runtime, graph, appstate, and events
- route app initialization through subsystem wrappers and add tests for the new lifecycle/config seams

## Validation
- `go test ./internal/app`
- `go test -count=1 ./internal/app`
- `go vet ./internal/app`
- `/Users/jonathan/go/bin/golangci-lint run --timeout 5m ./internal/app`
- pre-push changed-file checks for `./internal/app ./internal/findings`

Refs WriterInternal/cerebro#420

### DIFF
--- a/internal/app/app_subsystem_config.go
+++ b/internal/app/app_subsystem_config.go
@@ -57,19 +57,17 @@ type GraphConfig struct {
 }
 
 type AppStateConfig struct {
-	JobDatabaseURL       string
 	WarehouseBackend     string
 	WarehousePostgresDSN string
 }
 
 func (c AppStateConfig) DatabaseURL() string {
-	if dsn := strings.TrimSpace(c.JobDatabaseURL); dsn != "" {
-		return dsn
-	}
-	if strings.EqualFold(strings.TrimSpace(c.WarehouseBackend), "postgres") {
+	switch strings.ToLower(strings.TrimSpace(c.WarehouseBackend)) {
+	case "postgres", "snowflake":
 		return strings.TrimSpace(c.WarehousePostgresDSN)
+	default:
+		return ""
 	}
-	return ""
 }
 
 type EventConfig struct {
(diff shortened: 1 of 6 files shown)


internal/app/app_subsystem_config.go
```diff
@@ -0,0 +1,150 @@
...
+	MigrateLegacyActivityOnStart bool
+}
+
+type AppStateConfig struct {
+	JobDatabaseURL       string
+	WarehouseBackend     string
+	WarehousePostgresDSN string
+}
+
+func (c AppStateConfig) DatabaseURL() string {
```
josh-writer, line 64: AppStateConfig.DatabaseURL() now prefers JOB_DATABASE_URL and only falls back to WAREHOUSE_POSTGRES_DSN when WAREHOUSE_BACKEND=postgres. That diverges from the existing app-state contract in internal/app/app_state_postgres.go:40-50 and  internal/app/app_state_postgres_test.go:121-150, which uses WAREHOUSE_POSTGRES_DSN for both postgres and snowflake backends and explicitly ignores JOB_DATABASE_URL. Because the new lifecycle wrappers gate initAppStateDB / migrateAppState on this helper (internal/app/app_subsystems.go:21-39), a Snowflake deployment doing the Postgres app-state cutover will now skip Postgres app-state  init and migration entirely and silently stay on legacy Snowflake state.

josh-writer (changes requested):

internal/app/app_subsystem_config.go
jonathanhaaswriter: Aligned `AppStateConfig.DatabaseURL()` with the existing `appStateDatabaseURL()` contract so snowflake and postgres backends both use `WAREHOUSE_POSTGRES_DSN`, and added targeted tests to lock that behavior in.

josh-writer (approved):

Issue #261, opened 2026-04-14
Add unified endpoint software vulnerability correlation


jonathanhaaswriter:
## Summary
Add first-class support for unified endpoint software vulnerability correlation across endpoint inventory providers.

## Problem
Cerebro can ingest endpoint inventory and some provider-reported vulnerability data, but it does not yet offer a normalized, cross-provider view of:
- device -> installed software
- software -> CVEs
- CVEs -> patch/remediation target

This makes it hard to answer questions like "which endpoints have exploitable software that needs patching?" across multiple endpoint sources.

## Proposed scope
- Normalize installed software/application inventory across supported endpoint providers
- Normalize provider vulnerability records to a shared device/software/CVE model
- Add provider deduplication and freshness/confidence metadata
- Enrich CVEs with shared advisory metadata (severity, exploit-in-the-wild, prioritization)
- Expose query/API/reporting support for patch-target views by device, software, and CVE
- Add policy/report coverage for endpoint patch SLAs and exploited-in-the-wild prioritization

## Nice-to-have follow-ons
- Cover additional endpoint patch surfaces beyond apps/libraries where available, such as OS updates, browser extensions, agent versions, drivers, and firmware
- Add graph/report integration for endpoint vulnerability posture

## Acceptance criteria
- A user can identify affected endpoints for a given CVE
- A user can identify which software/version should be patched on each endpoint
- Results can be correlated across more than one endpoint provider without duplicate counting
- CVE prioritization includes exploitability signals where available

jonathanhaaswriter:
Closing as implemented for the current scope. Endpoint software vulnerability correlation now has normalized endpoint vulnerability storage/API support, canonical package/CVE graph links, and cross-source graph traversal support. No further UI/report work is planned for this slice.

PR #262, opened 2026-04-14
feat(endpoint): correlate software vulnerabilities across providers

jonathanhaaswriter:
## Summary
- materialize normalized `endpoints`, `endpoint_software_inventory`, and `vulnerabilities` tables from Kandji and SentinelOne provider data
- dedupe endpoint/software/CVE rows across providers and enrich them with advisory and KEV metadata for patch-priority queries
- refresh the normalized tables on startup, endpoint provider sync, and threat intel sync while preserving the existing workflow documentation comments

## Validation
- python3 scripts/devex.py run --mode changed --files internal/app/app.go internal/app/app_endpoint_vulnerabilities.go internal/app/app_init_phases.go internal/app/app_security_services.go internal/app/app_storage_graph.go internal/app/providers_registry.go internal/endpointvuln/service.go internal/endpointvuln/service_test.go internal/providers/hooks.go internal/providers/hooks_test.go
- pre-commit hooks on `git commit`
- pre-push checks on `git push`

Refs #261

### DIFF
--- a/internal/api/handlers_providers.go
+++ b/internal/api/handlers_providers.go
@@ -11,7 +11,17 @@ import (
 	"github.com/writer/cerebro/internal/providers"
 )
 
-// Provider endpoints
+// Provider endpoints.
+//
+// Endpoint software/CVE workflows generally use this control plane first:
+//   1. POST /api/v1/providers/{name}/sync to materialize provider rows
+//   2. GET /api/v1/providers/{name}/schema to discover table names and columns
+//   3. POST /api/v1/query (or GET /api/v1/assets/{table} for spot checks) to
+//      join device/agent, application, and vulnerability tables
+//
+// Today the provider layer preserves provider-native software names and versions.
+// That keeps ingest lossless, but cross-provider patch-target correlation usually
+// needs an extra normalization step above these endpoints.
 
 func (s *Server) listProviders(w http.ResponseWriter, r *http.Request) {
 	providerList := s.app.Providers.List()
--- a/internal/api/server_handlers_data.go
+++ b/internal/api/server_handlers_data.go
@@ -132,7 +132,16 @@ func parseLastSyncValue(value interface{}) time.Time {
 	return time.Time{}
 }
 
-// Query endpoints
+// Query endpoints.
+//
+// For endpoint patching/vulnerability flows, POST /api/v1/query is the main
+// data-plane surface after provider sync has materialized rows. Typical joins:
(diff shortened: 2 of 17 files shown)


josh-writer (changes requested):
•  [P1] Make normalized-table refresh atomic — internal/endpointvuln/service.go:353,977  
     Refresh() rewrites endpoints, endpoint_software_inventory, and vulnerabilities one table at a time, and replaceRows() does DELETE 
   FROM before reinserting rows without a transaction/swap. Any live query during a refresh can see mixed generations or an empty table,
    and a timeout/error after the delete leaves that table blank until the next successful refresh.

   •  [P1] Skip post-sync refresh after partial provider failures — internal/app/app_endpoint_vulnerabilities.go:65  
     endpointProviderSyncHook ignores the *providers.SyncResult. Kandji, SentinelOne, and CrowdStrike all report table-level failures 
   through result.Errors while still returning err == nil, so a sync with stale/failed source tables still triggers a normalized refresh
    and publishes mixed data.

   •  [P1] Propagate hook failures as real sync failures — internal/providers/hooks.go:27  
     WithSyncHook() appends hook errors to result.Errors but returns the original err. The direct scheduled-sync path only checks err, 
   so a failed normalized-table refresh can be reported as a successful scheduled provider sync while the derived tables stay stale.

   •  [P1] Honor cancellation for background refreshes — internal/app/app_endpoint_vulnerabilities.go:22, 
      internal/app/app_security_services.go:100  
     The threat-intel path now calls the refresher after sync, but the refresher strips cancellation with backgroundWorkContext() and 
   gives itself a new 2-minute timeout. App.Close() waits for the threat-intel goroutine, so shutdown can block far past the normal 
   shutdown budget if it lands during this refresh.

   •  [P1] Separate exploitability/public exploit from exploited-in-wild — internal/endpointvuln/service.go:606,1025  
     The new flow maps CrowdStrike exploit_available into ExploitedInWild, and also promotes advisory.Exploitable into the same flag. In
    this repo vulndb.LookupCVE() marks Exploitable for high-EPSS CVEs, so non-KEV vulns can be emitted as exploited_in_wild=true and 
   has_public_exploit=true, overstating active exploitation and inflating priority scoring.

   •  [P2] internal/providers/crowdstrike.go:344  
     queryResources() only returns the first page of IDs and never follows pagination, so larger CrowdStrike tenants will silently miss 
   hosts/detections beyond page 1.

PR #316, opened 2026-04-16
fix(app): guard insecure NATS TLS overrides

### LINKED ISSUE #268: enhancement: Add production guardrails for TLS InsecureSkipVerify configuration
## Summary

The `NATSJetStreamTLSInsecure` config flag is passed through to 10+ components (events router, consumer, replay, agents remote tools, tool publisher, graph writer lease, etc.) and sets `tls.Config{InsecureSkipVerify: true}`. There is no warning logged when this is enabled, no validation preventing it in production, and no way to audit whether it is active.

## Affected components

- `internal/events/router.go`, `consumer.go`, `replay.go`, `jetstream.go`
- `internal/agents/remote_tools.go`, `tool_publisher.go`
- `internal/app/app_init_agents.go`, `app_event_routing.go`, `app_init_core.go`
- `internal/app/graph_writer_lease_nats.go`
- `internal/app/app_stream_consumer_lifecycle.go`
- `internal/cli/ingest_replay.go`, `job_runtime.go`, `workload_scan.go`

## Impact

- TLS verification can be silently disabled in production
- MITM attacks on NATS connections would go undetected
- No audit trail for when/why insecure TLS was enabled

## Suggested fix

1. Emit a prominent warning log when `InsecureSkipVerify` is enabled
2. Consider requiring an explicit environment acknowledgment (e.g., `CEREBRO_ALLOW_INSECURE_TLS=true`) beyond just the NATS flag
3. Expose the TLS verification status in the health endpoint
4. Consider refusing `InsecureSkipVerify` entirely unless a `--dev` or `--local` mode flag is set

jonathanhaaswriter:
## Summary
- reject insecure NATS TLS skip-verify settings unless CEREBRO_ALLOW_INSECURE_TLS=true is set
- enforce the same guardrail across JetStream, remote tools, and tool publisher transports
- log an explicit startup warning and add regression coverage for the override path

Closes #268

### DIFF
--- a/internal/agents/remote_tools.go
+++ b/internal/agents/remote_tools.go
@@ -45,6 +45,7 @@ type RemoteToolProviderConfig struct {
 	TLSKeyFile            string
 	TLSServerName         string
 	TLSInsecureSkipVerify bool
+	AllowInsecureTLS      bool
 }
 
 type RemoteToolProvider struct {
@@ -368,7 +369,7 @@ func signerFromSeed(seed string) (string, func([]byte) ([]byte, error), error) {
 }
 
 func (c RemoteToolProviderConfig) tlsConfig() (*tls.Config, error) {
-	if c.TLSInsecureSkipVerify && !allowInsecureTLSOverride() {
+	if c.TLSInsecureSkipVerify && !allowInsecureTLSOverride(c.AllowInsecureTLS) {
 		return nil, fmt.Errorf("remote tools tls insecure skip verify requires CEREBRO_ALLOW_INSECURE_TLS=true")
 	}
 	tlsConfig := &tls.Config{
@@ -404,6 +405,6 @@ func (c RemoteToolProviderConfig) tlsConfig() (*tls.Config, error) {
 	return tlsConfig, nil
 }
 
-func allowInsecureTLSOverride() bool {
-	return strings.EqualFold(strings.TrimSpace(os.Getenv("CEREBRO_ALLOW_INSECURE_TLS")), "true")
+func allowInsecureTLSOverride(explicit bool) bool {
+	return explicit || strings.EqualFold(strings.TrimSpace(os.Getenv("CEREBRO_ALLOW_INSECURE_TLS")), "true")
 }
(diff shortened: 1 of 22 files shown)


josh-writer (changes requested):
[P1] Don’t treat the default alert router as an active NATS transport
   •  File: internal/app/app.go:326
   •  usesNATSTransportSettings() now counts AlertRouterEnabled, but startEventAlertRouting() bails out unless NATS_JETSTREAM_ENABLED is
       also true (internal/app/app_event_routing.go:19). Since ALERT_ROUTER_ENABLED defaults to true (internal/app/app_config.go:800), a
       config with only NATS_JETSTREAM_TLS_ENABLED=true and NATS_JETSTREAM_TLS_INSECURE_SKIP_VERIFY=true now fails validation 
      (internal/app/app_config_validation.go:433) even though no NATS client would be constructed. This is a concrete startup 
      regression.

   [P1] JetStream runtime override parsing is inconsistent with app boolean parsing
   •  File: internal/events/jetstream.go:949
   •  App-level validation/warnings use getEnvBool() (internal/app/app_env_helpers.go:211), which accepts true, 1, and yes. The new 
      allowInsecureTLSOverride() helper here only accepts literal "true". So CEREBRO_ALLOW_INSECURE_TLS=1 or yes will pass 
      cfg.Validate() and warning checks, but JetStream transport setup still fails at runtime in tlsConfig() 
      (internal/events/jetstream.go:914). This affects publisher/consumer/router/replay paths that all flow through JetStreamConfig.

   [P1] Agent transport override parsing has the same env-format regression
   •  File: internal/agents/remote_tools.go:408
   •  The agents-side helper also only accepts literal "true". That means existing deployments using CEREBRO_ALLOW_INSECURE_TLS=1 or yes
       can pass app-level validation, but RemoteToolProviderConfig.tlsConfig() (internal/agents/remote_tools.go:373) and 
      ToolPublisherConfig.tlsConfig() (internal/agents/tool_publisher.go:445) will still reject insecure TLS at runtime.

PR #330, opened 2026-04-16
enhance(remediation): explain unsupported terraform delivery

### LINKED ISSUE #271: enhancement: Terraform remediation has unimplemented delivery paths for several resource types
## Summary

`internal/remediation/terraform.go:461` returns a "not implemented" error for certain `(actionType, provider, resourceFamily)` combinations. Users attempting terraform-based remediation for unsupported resource types receive a generic "not implemented" error with no guidance on alternatives.

## Location

```go
// internal/remediation/terraform.go:461
return fmt.Errorf(
    "terraform delivery is not implemented for %s (provider=%s resource_family=%s)",
    actionType,
    firstNonEmpty(key.Provider, "unknown"),
    firstNonEmpty(key.ResourceFamily, "unknown"),
)
```

## Impact

- Users cannot use terraform remediation for certain provider/resource combinations
- The error message does not indicate which combinations ARE supported
- No fallback or alternative remediation path is suggested

## Suggested fix

1. Audit which `(actionType, provider, resourceFamily)` tuples hit the "not implemented" path
2. Implement the missing terraform delivery logic, or
3. At minimum, return a structured error that lists supported combinations and suggests manual remediation steps
4. Add a CLI command or API endpoint to enumerate supported remediation actions per resource type

jonathanhaaswriter:
## Summary
- enrich unsupported Terraform remediation errors with the supported provider/resource-family combinations for that action
- add explicit guidance to fall back to remote_apply or manual remediation when Terraform delivery is unavailable
- cover both provider mismatch and unsupported resource-family paths with regression tests

Closes #271

### DIFF
--- a/internal/remediation/terraform.go
+++ b/internal/remediation/terraform.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 
 	"github.com/writer/cerebro/internal/iacrender"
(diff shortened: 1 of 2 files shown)


josh-writer (changes requested):
•  [P3] Avoid recommending remote_apply when that path is impossible for the requested context — 
      internal/remediation/terraform.go:497  
     terraformSupportedContextsHelp() always appends consider remote_apply or manual remediation, but some unsupported contexts in this 
   PR have no viable remote_apply fallback. For example, enable_bucket_default_encryption is AWS-only in the catalog and only exposes 
   aws.s3.put_bucket_encryption (internal/remediation/catalog.go:123-160), so a GCP bucket will fail the provider precondition instead 
   of being remediable via remote_apply (internal/remediation/catalog_actions.go:127). The same applies to the unsupported 
   resource-family case for restrict_public_security_group_ingress (internal/remediation/catalog_actions.go:206). The new assertions in 
   internal/remediation/terraform_test.go:163 and :188 codify that misleading guidance.

PR #338, opened 2026-04-16
fix(api): sanitize sync response error details

### LINKED ISSUE #267: security: API error responses leak internal error details to clients
## Summary

Multiple API handlers in `internal/api/` return raw `fmt.Sprintf("...%v", err)` error messages directly to HTTP clients. These error messages can contain internal file paths, SQL query fragments, stack traces, or infrastructure details that should not be exposed externally.

## Examples

```
handlers_agents_runtime.go:77   — fmt.Sprintf("Error executing tool: %v", err)
server_handlers_graph_schema.go:42 — fmt.Sprintf("invalid node kind definition: %v", err)
server_handlers_platform_scan_audit.go:208 — fmt.Sprintf("failed to render scan audit package: %v", err)
server_handlers_sync.go:298,437,445,458 — account sync errors with raw %v
server_handlers_graph_intelligence.go:997 — fmt.Sprintf("load default graph mapping config: %v", err)
handlers_findings_compliance.go:749 — fmt.Sprintf("failed to render audit package: %v", err)
```

## Impact

- Internal error details (paths, SQL, config) could be disclosed to API callers
- Aids attackers in understanding internal architecture
- Potential compliance concern for SOC 2 / ISO 27001

## Suggested fix

- Return generic error messages to clients (e.g., "internal server error") for 5xx responses
- Log the detailed error server-side with a correlation ID
- Include the correlation ID in the client response for support tracing
- For 4xx validation errors, sanitize the error message to only include user-relevant details

jonathanhaaswriter:
## Summary
- sanitize relationship extraction failure reasons before returning sync responses
- sanitize AWS org account error details while logging the raw failures server-side
- add regression coverage for AWS, AWS org, and GCP sync sanitization paths

## Testing
- go test ./internal/api -run 'TestSync(AWS|AWSOrg|GCP)_(UsesRequestOptions|SanitizesRelationshipExtractionFailureReason|SanitizesAccountErrorsInResponse)$'

Fixes #267

### DIFF
--- a/internal/api/server_handlers_sync.go
+++ b/internal/api/server_handlers_sync.go
@@ -333,7 +333,7 @@ func (s *Server) syncAWS(w http.ResponseWriter, r *http.Request) {
 		"relationships_extracted": result.RelationshipsExtracted,
 	}
 	if result.RelationshipsSkippedReason != "" {
-		resp["relationships_skipped_reason"] = result.RelationshipsSkippedReason
+		resp["relationships_skipped_reason"] = s.sanitizeSyncRelationshipsSkippedReason("aws", result.RelationshipsSkippedReason)
 	}
 	if graphUpdate := result.GraphUpdate; graphUpdate != nil {
 		resp["graph_update"] = graphUpdate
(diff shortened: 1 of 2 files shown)


josh-writer (changes requested):
•  [P1] Sanitize table-level sync errors before returning results — internal/api/server_handlers_sync.go:332, 
      internal/api/server_handlers_sync.go:508, internal/api/server_handlers_sync.go:736  
     This patch sanitizes relationships_skipped_reason and account_errors, but all three handlers still serialize result.Results 
   verbatim. nativesync.SyncResult.Error is populated from raw err.Error() in the sync engines (for example internal/sync/engine.go:405 
   and internal/sync/gcp.go:224), so a failing table can still leak backend details like file paths, bucket URIs, ARNs, or internal 
   error strings in the API response. The new tests only cover the new top-level sanitizers 
   (internal/api/server_handlers_sync_test.go:373, internal/api/server_handlers_sync_test.go:885, 
   internal/api/server_handlers_sync_test.go:1013) and would not catch results[*].Error leakage.

PR #340, opened 2026-04-16
fix(app): order snowflake findings initialization

### LINKED ISSUE #321: bug: initFindings and initSnowflakeFindings race in phase 2a — silent data loss when SnowflakeFindings is nil
## Summary

`initFindings()` and `initSnowflakeFindings()` both run concurrently in phase 2a. `initSnowflakeFindings` reads `a.SnowflakeFindings` which is written by `initFindings`. If `initSnowflakeFindings` runs first, `a.SnowflakeFindings` is nil, and the method returns early — silently skipping the Snowflake findings load.

## Evidence

`app_init_phases.go` phase 2a concurrent block:
```go
initOnlySubsystem("findings", func(context.Context) { a.initFindings() }),
// ...
initOnlySubsystem("snowflake_findings", func(taskCtx context.Context) { a.initSnowflakeFindings(taskCtx) }),
```

`initSnowflakeFindings` reads the field:
```go
func (a *App) initSnowflakeFindings(ctx context.Context) {
    if a.Snowflake == nil || a.SnowflakeFindings == nil {  // ← reads before initFindings writes it
        return
    }
    if err := a.SnowflakeFindings.Load(ctx); err != nil { ... }
}
```

`initFindings` writes the field (near end of method):
```go
a.SnowflakeFindings = snowflakeStore
```

## Impact

- **Silent data loss**: Historical findings from Snowflake are never loaded into the in-memory store
- **Race is timing-dependent**: In practice, `initFindings` usually wins because it runs first in the goroutine pool, but there's no ordering guarantee
- **No error or warning**: The early return produces no log message — the app starts with empty findings history

## Fix options

1. Move `initSnowflakeFindings` out of the concurrent block and run it after `initFindings` completes
2. Split the dependency: have `initFindings` return the snowflake store reference, and pass it explicitly to `initSnowflakeFindings`
3. Move the `SnowflakeFindings.Load()` call into `initFindings` itself

## Related

- #294 — Data race in initPhase2b (agents/remediation) — same class of concurrent init race

jonathanhaaswriter:
## Summary
- load snowflake-backed findings from within the findings init path using the phase context
- remove the separate concurrent snowflake_findings task so startup ordering is deterministic
- add regression coverage for loading the snowflake findings store during findings initialization

## Testing
- go test ./internal/app -run 'TestInitFindings_'

Fixes #321


jonathanhaaswriter:
## Review: fix(app): order snowflake findings initialization

Fix is correct but **needs rebase** — the current branch refactored `initPhase2a` to use `initOnlySubsystem(...)` wrappers. This PR targets the old `{name: ..., run: ...}` format and will conflict.

PR #349, opened 2026-04-16
fix(app): remove dead compliance wiring

### LINKED ISSUE #323: enhancement: Compliance field on App is declared but never assigned — dead code
## Summary

The `Compliance` field on the App struct is declared but never assigned a value anywhere in the codebase. It is only read in `app_service_groups.go` where it is always nil.

## Evidence

Declaration in `app.go`:
```go
Compliance *compliance.ComplianceReport
```

Read in `app_service_groups.go`:
```go
// SecurityServices() does NOT include Compliance
```

Grepping all non-test files in `internal/app/` for `a.Compliance` or `app.Compliance`:
- **Zero writes** — no `a.Compliance =` assignment anywhere
- **Zero init** — no `initCompliance` sets this field

The `initCompliance()` method in `app_security_services.go` exists but only logs:
```go
func (a *App) initCompliance() {
    // This method exists but does NOT set a.Compliance
    a.Logger.Info("compliance report service initialized")
}
```

## Impact

- Dead field consuming memory on every App instance
- Misleading API — consumers of `SecurityServices()` or the API layer may check `Compliance != nil` and always get false
- The `compliance` package itself is functional (has tests) — the wiring is just missing

## Fix

Either:
1. Wire `initCompliance` to actually set `a.Compliance = compliance.NewComplianceReport(...)`
2. Remove the field if compliance reporting is not yet implemented at the app level

jonathanhaaswriter:
## Summary
- remove the unused App.Compliance field and the matching SecurityServices field
- drop the no-op compliance init task and initCompliance helper
- keep the app wiring aligned with the on-demand compliance reporting model

## Testing
- go test ./internal/app

Fixes #323

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -49,7 +49,6 @@ import (
 	"github.com/writer/cerebro/internal/attackpath"
 	"github.com/writer/cerebro/internal/auth"
 	"github.com/writer/cerebro/internal/cache"
-	"github.com/writer/cerebro/internal/compliance"
 	"github.com/writer/cerebro/internal/dspm"
 	"github.com/writer/cerebro/internal/events"
 	"github.com/writer/cerebro/internal/executionstore"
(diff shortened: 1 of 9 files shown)


jonathanhaaswriter:
## Review: fix(app): remove dead compliance wiring

Removal is correct but **needs rebase** — same `initOnlySubsystem` format conflict as #340.

PR #363, opened 2026-04-16
fix(sync): allow azure schema evolution

### LINKED ISSUE #342: bug: Azure sync engine ensureTable uses AddMissingColumns:false — schema can never evolve
## Summary

The Azure sync engine's `ensureTable` helper passes `AddMissingColumns: false` to the Snowflake table operations. This means if a new column is added to an Azure resource type (e.g., Azure adds a new property), the sync engine will silently drop the new field because the existing table schema won't be updated.

## Evidence

In `internal/sync/azure.go`, the table ensure call uses:
```go
tableops.EnsureTableOptions{
    AddMissingColumns: false,  // schema frozen after creation
}
```

Compare with the AWS sync engine which uses `AddMissingColumns: true`, allowing schema evolution.

## Additional Azure sync gaps

1. **No CDC event emission**: Azure sync doesn't emit change events after syncing, unlike AWS which emits CDC events for the graph builder
2. **No generation tracking**: Azure sync doesn't track sync generations or staleness, so there's no way to detect stale Azure data
3. **Missing context cancellation checks**: 15+ Azure pagination loops don't check `ctx.Done()` between pages

## Impact

- New Azure resource properties are silently lost during sync
- The graph builder never sees new Azure fields, leading to incomplete security analysis
- Azure data staleness is undetectable

## Related

- #308 — Provider pagination without loop protection
- #289 — DB schema migrations use CREATE IF NOT EXISTS with no versioning

jonathanhaaswriter:
## Summary
- allow Azure ensureTable calls to add missing columns when provider schemas grow
- extend the static ensure-table regression to cover Azure as well

## Testing
- go test ./internal/sync

Fixes #342

### DIFF
--- a/internal/sync/azure.go
+++ b/internal/sync/azure.go
@@ -586,7 +586,7 @@ func (e *AzureSyncEngine) validateTable(ctx context.Context, table AzureTableSpe
 
 func (e *AzureSyncEngine) ensureTable(ctx context.Context, table string, columns []string) error {
 	return tableops.EnsureVariantTable(ctx, e.sf, table, columns, tableops.EnsureVariantTableOptions{
-		AddMissingColumns: false,
+		AddMissingColumns: true,
 	})
 }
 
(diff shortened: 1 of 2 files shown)


jonathanhaaswriter:
## Review: fix(sync): allow azure schema evolution

**Incomplete parity fix.**

AWS/GCP/K8s sync engines all pass three `EnsureTableOptions` flags (`AddMissingColumns`, `IgnoreLookupError`, `IgnoreAddColumnErrors`). This PR only flips `AddMissingColumns` for Azure — missing the two error tolerance flags means Azure sync will hard-fail on transient DDL errors while other providers continue.

Add `IgnoreLookupError: true` and `IgnoreAddColumnErrors: true` to match, and update the static assertion.

PR #369, opened 2026-04-16
fix(events): recover panicking consumer handlers

### LINKED ISSUE #362: bug: JetStream consumer pipeline has no panic recovery — panicking handler crashes entire process
## Summary

The JetStream consumer pipeline in `internal/events/consumer_pipeline.go` invokes the user-provided message handler inside a goroutine with no `recover()` wrapper. A panicking handler (e.g., nil pointer dereference in event processing logic) crashes the entire process instead of being treated as a handler failure and nacked for retry.

Additionally, a manually-constructed `Consumer` struct with a nil handler causes an unrecoverable nil-function panic in the pipeline worker goroutine.

## Evidence (TDD — observed in `events_tdd_test.go`)

**Test 1: `TestTDD_ConsumerProcessBatchHandlerPanics`**
- Handler: `func(...) error { panic("boom") }`
- Result: Process crashes with `panic: boom` from the pipeline transform goroutine
- Expected: Handler panic caught, message nacked, consumer continues

**Test 2: `TestTDD_ConsumerProcessBatchNilHandler`**
- Consumer constructed with nil handler
- Result: Nil function call panic in `transformDecodedBatchMessage`
- Note: `NewJetStreamConsumer` constructor guards against nil handler, but the `Consumer` struct is exported and can be constructed directly

## Impact

- A single malformed event that triggers a panic in the handler (e.g., nil field dereference) takes down the entire consumer process
- In production: all event processing stops, requiring a restart
- Events that caused the panic will be redelivered on restart, potentially causing a crash loop

## Fix

Add `recover()` in the pipeline transform goroutine:
```go
func (c *Consumer) transformDecodedBatchMessage(ctx context.Context, msg DecodedMessage) pipelineResult {
    defer func() {
        if r := recover(); r != nil {
            // Log panic, return error result so message is nacked
        }
    }()
    // ... existing handler invocation
}
```

## Related

- #272 — Goroutine panic recovery
- #336 — Event consumer deadlock risk

jonathanhaaswriter:
## Summary
- recover panics and nil handlers before invoking event consumer callbacks
- route recovered handler failures through the existing retry/nak path in both single-message and batch processing
- add regression tests for panicking and missing handlers in the batch pipeline

## Validation
- go test ./internal/events

Closes #362

### DIFF
--- a/internal/events/consumer.go
+++ b/internal/events/consumer.go
@@ -40,6 +40,8 @@ const (
 	defaultConsumerLagRefreshInterval  = 15 * time.Second
 )
 
+var errConsumerHandlerRequired = errors.New("consumer handler is required")
+
 type ConsumerConfig struct {
 	URLs                []string
 	Stream              string
@@ -169,7 +171,7 @@ type ConsumerHealthSnapshot struct {
 
 func NewJetStreamConsumer(cfg ConsumerConfig, logger *slog.Logger, handler EventHandler) (*Consumer, error) {
 	if handler == nil {
-		return nil, errors.New("consumer handler is required")
+		return nil, errConsumerHandlerRequired
 	}
 	config := cfg.withDefaults()
 	if err := config.validate(); err != nil {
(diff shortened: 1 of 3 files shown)


jonathanhaaswriter:
## Review: fix(events): recover panicking consumer handlers

**2 items to address before merge:**

### 1. Missing `debug.Stack()` on panic recovery
The project's established panic-recovery pattern (see `jobs/worker.go`, `scheduler/scheduler.go`) captures `debug.Stack()` alongside the recovered value. This PR only captures the panic value — in production this makes panics very hard to debug. Add:
```go
fmt.Errorf("consumer handler panic: %v\n%s", recovered, debug.Stack())
```

### 2. No test for single-message handler path
Both new tests only exercise `processBatch`. A panic-recovery test for `handleDecodedMessage` should be added to verify both code paths.

**Fix the stack trace capture and add the single-message test, then this is ready to merge.**

PR #373, opened 2026-04-16
fix(app): detect missing snowflake tables by code

### LINKED ISSUE #353: bug: isMissingSnowflakeTableErr uses string matching — can match non-table errors
## Summary

The migration guard function `isMissingSnowflakeTableErr` in `internal/app/app_state_postgres.go` uses `strings.Contains` on the lowercased error message to detect missing Snowflake tables. This is fragile and can match unrelated errors.

## Evidence

```go
func isMissingSnowflakeTableErr(err error) bool {
    if err == nil {
        return false
    }
    message := strings.ToLower(err.Error())
    return strings.Contains(message, "does not exist") ||
        strings.Contains(message, "unknown table") ||
        strings.Contains(message, "not exist")
}
```

**Problems:**
1. `"does not exist"` matches column-not-exist, schema-not-exist, role-not-exist, etc.
2. `"not exist"` matches any error whose wrapped chain contains "not exist" — e.g., "user does not exist", "database does not exist"
3. No check for Snowflake-specific error codes (Snowflake returns structured error codes like `002003` for missing objects)

## Impact

- A permissions error containing "role does not exist" would be silently swallowed, making the migration appear successful when it actually failed
- A typo in the Snowflake database/schema config would be silently swallowed instead of surfacing to the operator
- Once Snowflake is decommissioned, ANY error from a stale connection string would potentially match

## Suggested fix

Use Snowflake driver error codes instead of string matching:
```go
func isMissingSnowflakeTableErr(err error) bool {
    var sfErr *gosnowflake.SnowflakeError
    if errors.As(err, &sfErr) {
        // 002003 = SQL compilation error (object does not exist)
        return sfErr.Number == 2003
    }
    return false
}
```

## Related

- #293 — SQLite schema migration uses error string matching
- #352 — No migration-complete flag

jonathanhaaswriter:
## Summary
- require a wrapped `gosnowflake.SnowflakeError` when deciding whether to skip legacy app-state migrations
- only treat Snowflake object-does-not-exist errors as skippable and keep authorization-style failures visible
- add regression coverage for wrapped driver errors and generic string errors

## Validation
- go test ./internal/app

Closes #353

### DIFF
--- a/internal/app/app_state_postgres.go
+++ b/internal/app/app_state_postgres.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	_ "github.com/lib/pq"
+	sf "github.com/snowflakedb/gosnowflake"
 
 	"github.com/writer/cerebro/internal/agents"
 	"github.com/writer/cerebro/internal/appstate"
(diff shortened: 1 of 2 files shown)


jonathanhaaswriter:
Closing as superseded by #366. #366 keeps the Snowflake error narrowing while also restricting skips to known app-state tables, which is the safer variant to keep.

PR #376, opened 2026-04-16
fix(policy): preserve empty-description violations

### LINKED ISSUE #364: bug: Policy engine silently drops findings when Description is empty — checkAssetViolation conflates violation signal with message
## Summary

`checkAssetViolation()` in `internal/policy/engine.go` returns `p.Description` as the violation result. `EvaluateAsset()` then checks `violation != ""` to decide whether to create a finding. This means a policy with an empty `Description` field **silently produces zero findings** even when all conditions match.

## Evidence (TDD — failing test in `tdd_edge_cases_test.go`)

```
Test: TestTDDPolicyEmptyDescriptionSilentlyDropsFindings
  Two identical policies, same conditions, both match the asset.
  Policy "no-desc" has Description: "" → produces NO finding
  Policy "has-desc" has Description: "something" → produces 1 finding
  Expected: 2 findings
  Got: 1 finding
```

Root cause in `engine.go`:
```go
func (e *Engine) checkAssetViolation(p *Policy, asset map[string]interface{}) string {
    if len(p.Conditions) == 0 { return "" }
    if normalizeConditionFormat(p.ConditionFormat) != ConditionFormatCEL { return "" }
    if !e.evaluateCELConditions(p, asset) { return "" }
    return p.Description  // ← empty Description = no violation, even though conditions matched
}

// In EvaluateAsset:
if violation := e.checkAssetViolation(p, asset); violation != "" {
    // Only creates finding when violation is non-empty
}
```

## Impact

- Any policy loaded without a Description field (e.g., from YAML with missing `description:` key) silently evaluates but never produces findings
- `AddPolicy()` doesn't validate Description is non-empty (only `ValidatePolicy` does, but it's not called by `AddPolicy`)
- Programmatically-added policies (tests, SDK, dynamic policy loading) are especially vulnerable
- No error or warning is logged — the policy appears to be working but silently does nothing

## Fix

Option 1: Return a sentinel value from `checkAssetViolation` instead of using empty string as "no violation":
```go
func (e *Engine) checkAssetViolation(p *Policy, asset map[string]interface{}) (string, bool) {
    // ... 
    return p.Description, true  // description can be empty, bool indicates violation
}
```

Option 2: Default Description to policy Name when empty:
```go
violation := p.Description
if violation == "" { violation = p.Name }
return violation
```

## Related

- This is NOT caught by `ValidatePolicy`/`validateLoadedPolicy` because those ARE called during file-based loading (which requires Description). But `AddPolicy()` bypasses that validation.

jonathanhaaswriter:
## Summary
- separate violation matching from the optional description payload in policy evaluation
- keep versioning candidate evaluation aligned with that behavior
- add regression coverage for empty-description policies in both evaluation paths

## Testing
- go test ./internal/policy

Fixes #364

### DIFF
--- a/internal/policy/empty_description_regression_test.go
+++ b/internal/policy/empty_description_regression_test.go
@@ -1,0 +1,68 @@
+package policy
+
+import (
+	"context"
+	"testing"
+)
+
+func TestEvaluateAssetReportsViolationsWithEmptyDescription(t *testing.T) {
+	engine := NewEngine()
+	engine.AddPolicy(&Policy{
+		ID:              "no-desc",
+		Name:            "No description policy",
+		Effect:          "forbid",
+		Resource:        "aws::s3::bucket",
+		ConditionFormat: ConditionFormatCEL,
+		Conditions:      []string{`cmp_eq(path(resource, "public"), true)`},
+		Severity:        "high",
+	})
+	engine.AddPolicy(&Policy{
+		ID:              "has-desc",
+		Name:            "Has description policy",
+		Description:     "Public bucket detected",
+		Effect:          "forbid",
+		Resource:        "aws::s3::bucket",
+		ConditionFormat: ConditionFormatCEL,
+		Conditions:      []string{`cmp_eq(path(resource, "public"), true)`},
+		Severity:        "high",
+	})
+
+	findings, err := engine.EvaluateAsset(context.Background(), map[string]interface{}{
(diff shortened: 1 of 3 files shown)


jonathanhaaswriter:
Closing as superseded by #384, which covers the same empty-description regression with broader finding-generation behavior and stronger regression tests.

PR #385, opened 2026-04-16
fix(appstate): serialize schema initialization

### LINKED ISSUE #365: bug: AppState/Agent repositories have no EnsureSchema synchronization — concurrent access races on DDL
## Summary

All appstate repositories (`AuditRepository`, `PolicyHistoryRepository`, `RiskEngineStateRepository`) and the `PostgresSessionStore` in `internal/agents/` call `EnsureSchema()` inside every read/write operation. `EnsureSchema()` executes DDL (`CREATE TABLE IF NOT EXISTS`, `CREATE INDEX IF NOT EXISTS`) without any `sync.Once` or mutex protection, creating a race where concurrent goroutines execute DDL simultaneously.

## Evidence (TDD — failing tests with `-race`)

```
Test: TestTDDAuditRepository_ConcurrentLogRace — FAIL
  20 goroutines call Log() concurrently
  Multiple goroutines hit: "SQL logic error: no such table: cerebro_audit_log"
  Expected 20 entries, got 2

Test: TestTDDRiskEngineStateRepository_ConcurrentSaveLoadRace — FAIL
  20 goroutines do concurrent SaveSnapshot/LoadSnapshot
  Multiple goroutines hit: "SQL logic error: no such table: cerebro_risk_engine_state"

Test: TestTDDSessionStore_ConcurrentCreatesRace — FAIL
  10 goroutines call Save() concurrently
  Multiple goroutines hit: "SQL logic error: no such table: cerebro_agent_sessions"
```

## Root cause

Each repository's `EnsureSchema()` follows this pattern:
```go
func (r *AuditRepository) EnsureSchema(ctx context.Context) error {
    _, err := r.db.ExecContext(ctx, `CREATE TABLE IF NOT EXISTS cerebro_audit_log (...)`)
    if err != nil { return err }
    _, err = r.db.ExecContext(ctx, `CREATE INDEX IF NOT EXISTS ...`)
    return err
}
```

And every write method calls it:
```go
func (r *AuditRepository) Log(ctx context.Context, entry AuditEntry) error {
    if err := r.EnsureSchema(ctx); err != nil { return err }
    // ... INSERT
}
```

When goroutine A's `CREATE TABLE` is still executing, goroutine B's `INSERT` runs and sees no table yet → error.

## Impact

- On first boot with a fresh database, any concurrent access to appstate tables will race
- The migration code (`migrateAppState`) calls `EnsureSchema` in series, but if API handlers or background jobs start before migration completes, they can race
- SQLite is especially vulnerable because DDL is not concurrent-safe without WAL mode
- Postgres can also race under high connection count (multiple `CREATE TABLE IF NOT EXISTS` executing simultaneously)

## Fix

Guard `EnsureSchema` with `sync.Once` in each repository:
```go
type AuditRepository struct {
    db         *sql.DB
    schemaOnce sync.Once
    schemaErr  error
}

func (r *AuditRepository) EnsureSchema(ctx context.Context) error {
    r.schemaOnce.Do(func() {
        r.schemaErr = r.ensureSchemaImpl(ctx)
    })
    return r.schemaErr
}
```

## Related

- #289 — DB schema migrations use CREATE IF NOT EXISTS with no versioning
- #351 — migrateAppState runs on every startup

jonathanhaaswriter:
## Summary
- guard appstate and session-store EnsureSchema paths with one-time initialization mutexes
- only mark schema setup complete after successful DDL so transient failures can retry
- add concurrent first-use regressions for audit, policy history, risk state, and session stores

Closes #365

### DIFF
--- a/internal/agents/session_store_postgres.go
+++ b/internal/agents/session_store_postgres.go
@@ -6,14 +6,17 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 )
 
 const postgresSessionTable = "cerebro_agent_sessions"
 
 type PostgresSessionStore struct {
-	db         *sql.DB
-	rewriteSQL func(string) string
+	db          *sql.DB
+	rewriteSQL  func(string) string
+	schemaMu    sync.Mutex
+	schemaReady bool
 }
 
 func NewPostgresSessionStore(db *sql.DB) *PostgresSessionStore {
@@ -24,6 +27,13 @@ func (s *PostgresSessionStore) EnsureSchema(ctx context.Context) error {
 	if s == nil || s.db == nil {
 		return fmt.Errorf("postgres session store is not initialized")
 	}
+
+	s.schemaMu.Lock()
+	defer s.schemaMu.Unlock()
+	if s.schemaReady {
+		return nil
+	}
+
 	_, err := s.db.ExecContext(ctx, s.q(`
 CREATE TABLE IF NOT EXISTS `+postgresSessionTable+` (
 	id TEXT PRIMARY KEY,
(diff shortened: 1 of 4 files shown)


jonathanhaaswriter:
Closing as superseded by #379. #379 covers the same schema-initialization race while also handling waiter context cancellation and broader first-use concurrency paths.

PR #462, opened 2026-04-23
feat(report): include graph evidence in finding summaries

jonathanhaaswriter:
## Summary
- enrich the built-in `finding-summary` report with bounded graph evidence for top finding resources
- plumb the graph query store through the bootstrap report service so report runs can compose persisted findings with local graph neighborhoods
- add focused report and bootstrap coverage for resource-root selection, bounded graph reads, and unconfigured-graph fallback

## Testing
- make verify
- go test ./internal/reports ./internal/bootstrap
- local graph-backed finding-summary report demo

PR #461, opened 2026-04-23
feat(report): add durable report run foundation

jonathanhaaswriter:
## Summary
- add a durable report-run foundation with built-in report definitions, persisted report runs, and bootstrap REST/Connect endpoints
- add the first built-in `finding-summary` report that materializes one runtime-scoped summary of persisted findings by severity, status, and rule
- extend the Postgres finding store with runtime-scoped finding reads so reports can compose over persisted findings instead of introducing a separate compliance module

## Testing
- make verify
- go test ./internal/reports ./internal/bootstrap ./internal/statestore/postgres ./internal/findings
- local finding-summary report demo

PR #515, opened 2026-05-07
feat: add source runtime flow telemetry

jonathanhaaswriter:
## Summary
- add runtime_id filtering for source-runtime listing and orchestrator runs
- emit structured JSON telemetry spans/events for orchestrator iterations, per-runtime work, sync pages, and commits
- propagate trace_id/span_id onto emitted source events and include runtime result counters

## Validation
- go test ./cmd/cerebro ./internal/sourceruntime ./internal/statestore/postgres ./internal/bootstrap ./internal/telemetry -count=1
- make verify

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -67,6 +67,10 @@ paths:
     get:
       summary: List source runtime configurations
       parameters:
+        - name: runtime_id
+          in: query
+          schema:
+            type: string
         - name: tenant_id
           in: query
           schema:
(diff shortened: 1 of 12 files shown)


jonathanhaaswriter:
This is the runtime_id-capable app dependency for WriterInternal/cerebro#461. All checks are passing and review threads are resolved; requesting review so we can cut v2.1.0 and unblock sec-dev rollout tracking in WriterInternal/cerebro#462.

PR #556, opened 2026-05-12
fix(grc): link vulnerabilities to targets

jonathanhaaswriter:
## Summary
- Add GRC target entities for vulnerability target IDs.
- Link GRC vulnerability targets to vulnerabilities, affected packages, and source integrations.
- Reuse GRC integration source projection helpers and assert target/integration field normalization.

## Tests
- go test ./internal/sourceprojection ./sources/grc
- make verify


prestrike00 (approved):
LGTM

PR #560, opened 2026-05-13
fix(findings): rework SentinelOne infection severity and guard future stale timestamps

jonathanhaaswriter:
## What

Reworks active-infection severity scoring and adds a future-timestamp guard to the stale-agent rule for SentinelOne graph evaluation.

## Why

Findings observed in `cerebro-sec-dev` showed:

- An infected host with multiple LOLBin threats (`control.exe`, `mshta.exe`, `secedit.exe`, `unlodctr.exe`) rated `HIGH`. Severity flattened to HIGH whenever there was only a single threat row, even when that threat was fileless or stuck in a failed mitigation state for days.
- Stale-agent findings churning for runtimes whose `last_active_date` is reported in the future (clock skew on small Linux fleets).

### `sentinelOneGraphInfectionSeverity`
- `CRITICAL` if any threat is fileless.
- `CRITICAL` if `>=2` threats AND at least one is unresolved.
- `CRITICAL` if any threat has been in `failed` / `requires_action` / `not_mitigated` mitigation status for `>24h`.
- `HIGH` otherwise.

`now` is threaded through so the aging window is testable and consistent with the rule's `EvaluateRows` clock.

### `sentinelOneAgentStaleGraphRule.EvaluateRows`
- Skips rows where `last_active_date.After(now)`.

### Tests
- New cases for each severity branch (fileless / multi-unresolved / aged failed mitigation / single resolved => HIGH).
- New case asserting future-dated `last_active_date` skipped by stale rule.

## Validation

- `go test ./internal/findings/...` passes.
- `make verify` passes.

## Out of scope

- Per-host stale-agent breakdown (currently still groups by SentinelOne site).
- OS vs binary-family mismatch detection (separate proposal for a new rule).

### DIFF
--- a/internal/findings/sentinelone_graph_rule.go
+++ b/internal/findings/sentinelone_graph_rule.go
@@ -215,7 +215,7 @@ func (r *sentinelOneEndpointActiveInfectionGraphRule) buildFinding(runtime *cere
 		RuntimeID:         strings.TrimSpace(runtime.GetId()),
 		RuleID:            r.definition.ID,
 		Title:             r.definition.Name,
-		Severity:          sentinelOneGraphInfectionSeverity(agentAttrs, threats),
+		Severity:          sentinelOneGraphInfectionSeverity(agentAttrs, threats, now),
 		Status:            r.definition.Status,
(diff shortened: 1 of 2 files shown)


jonathanhaaswriter:
Closing: opened in error, content not approved for public review.

PR #582, opened 2026-05-13
feat(cosmo): support message runtime pagination

jonathanhaaswriter:
## Summary
- allow Cosmo message reads without a ticket_id so runtime config can ingest all messages
- pass offset/limit to Cosmo and preserve cursor pagination for message events
- update tests for optional ticket_id pagination

## Dependency
Requires WriterInternal/cosmo PR for offset-capable all-message export before the message runtime is enabled.

## Tests
- go test ./sources/cosmo ./internal/sourceregistry -count=1 -v
- pre-push verify hook passed


jonathanhaaswriter:
Cosmo upstream dependency is now merged and deployed: WriterInternal/cosmo#275, Deploy Cloudflare Worker run 25833574964. Ticketless message reads are available before this connector PR is merged/released.

PR #594, opened 2026-05-19
feat(auth): emit access audit telemetry

jonathanhaaswriter:
## Summary
- emit structured `cerebro.api.access` telemetry for authenticated non-public Cerebro API requests
- log safe audit fields: outcome, status, method, sanitized route, duration, tenant/principal/auth mode, credential/client IDs, and Connect procedure
- classify downstream 401/403 responses as access denials while avoiding auth headers, request bodies, query strings, and raw resource paths

## Validation
- `make verify`

### DIFF
--- a/internal/bootstrap/app_test.go
+++ b/internal/bootstrap/app_test.go
@@ -1803,13 +1803,43 @@ func TestAuthMiddlewareSkipsAccessAuditForPublicRoutes(t *testing.T) {
 }
 
 func TestAccessAuditOutcomeClassifiesDownstreamAuthorizationFailures(t *testing.T) {
-	outcome, reason := accessAuditOutcome(http.StatusForbidden, "allowed", "")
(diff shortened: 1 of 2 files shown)


jonathanhaaswriter:
Deep follow-up pass:

- Added gRPC/HTTP2 regression coverage for scoped Connect procedure denials that return HTTP 200 with `permission_denied` trailers; audit outcome is now `denied` with `connect_code=permission_denied`.
- Sanitized unknown Connect procedure names to `unknown` to avoid raw path/cardinality leakage.
- Normalized `remote_addr` to `remote_ip` to drop per-connection ports.
- Ran a tmux-hosted local server experiment against `/health`, unauthenticated `/sources`, authenticated `/sources`, forbidden tenant access, and an unknown Connect procedure. Logs contained sanitized routes and did **not** contain the test bearer token, query secret, or unknown path segment.

Validation:
- `make verify`

PR #623, opened 2026-05-22
fix(vulndb): support windows release builds

jonathanhaaswriter:
## Summary
- Move file-store advisory DB locking behind OS-specific helpers so GoReleaser Windows targets compile.
- Preserve flock-based locking on non-Windows platforms; Windows uses the existing in-process mutex only.

## Validation
- go test ./internal/vulndb -count=1 -v
- GOOS=windows GOARCH=amd64 go test -c ./internal/vulndb -o /tmp/cerebro-vulndb-windows.test.exe
- make verify

### DIFF
--- a/internal/vulndb/filestore.go
+++ b/internal/vulndb/filestore.go
@@ -9,7 +9,6 @@ import (
 	"path/filepath"
 	"sort"
 	"sync"
-	"syscall"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


jonathanhaaswriter:
Closing as superseded by 3f8a1815cc479a65520b93d9b0f5dfe7022c1b1a on main, which implements platform-specific FileStore locking and is included in v2.1.54.

PR #625, opened 2026-05-22
ci: dispatch internal deploy on release

jonathanhaaswriter:
## Summary

This adds the public `writer/cerebro` half of the release-to-deploy handoff.

- Exposes the GHCR multi-arch manifest digest from the release job.
- Replaces the old `gh workflow run propose-image-tag.yml` matrix for both `sec-dev` and `go-prod` with a single signed release dispatch to `WriterInternal/cerebro`.
- Sends `image_tag`, `image_digest`, release URL, source repo/ref, `target_environment: sec-dev`, and `apply_mode: direct_push`.

## Why

The previous release handler could only ask the internal repo to open review-gated PRs, and it asked for `sec-dev` and `go-prod` at the same time. The new shape lets the internal repo automate the safe environment first (`sec-dev`) while keeping production behind the internal review/environment-gated promotion path.

## Internal dependency

Depends on the internal handler PR: https://github.com/WriterInternal/cerebro/pull/577

Merge order should be internal first, then this public PR. If this lands first, the dispatch can fire before the internal repo has a matching `repository_dispatch` handler.

## Validation

- `actionlint .github/workflows/release.yml`
- YAML parse for the modified workflow
- `git diff --check`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,6 +72,8 @@ jobs:
   ghcr-publish:
     runs-on: ubuntu-latest
     needs: [resolve-tag, goreleaser]
+    outputs:
+      digest: ${{ steps.manifest.outputs.digest }}
     env:
       RELEASE_TAG: ${{ needs.resolve-tag.outputs.tag }}
     steps:
@@ -137,6 +139,7 @@ jobs:
             --push .
 
       - name: Create and push multi-arch manifests
+        id: manifest
         shell: bash
         env:
           IMAGE_BASE: ghcr.io/${{ github.repository }}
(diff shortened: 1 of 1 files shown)


jonathanhaaswriter:
Addressed Droid's silent-failure finding in b58530af: the release workflow now fetches WriterInternal/cerebro's default-branch handler before dispatch and fails closed if `repository_dispatch` / `cerebro-release-published` is not installed. After dispatch, it waits up to 120s for a downstream `propose-image-tag.yml` `repository_dispatch` run to appear and fails if no run is created, so a missing/disabled handler can no longer report success.

PR #631, opened 2026-05-23
feat(sources): add aurelius S3-backed source for image scan telemetry

prestrike00:
### Why

Adds a new internal-telemetry source to Cerebro that ingests NDJSON
archives from S3 on a polling schedule and emits CloudEvents into the
append log. Pattern follows the existing `kolide` and `sentinelone`
S3-backed sources already in this repo.

### What changed

- `sources/<source>/catalog.yaml` — emitted kinds + event_contracts
  with `required_attributes` and `payload_fields` for each family.
- `sources/<source>/source.go` — `Source` implementation: lists keyed
  prefixes, resumes from cursor, handles `.ndjson` and `.ndjson.gz`,
  dispatches via `FamilyEngine` to per-family parsers; typed sentinel
  errors for misconfig.
- `sources/<source>/source_test.go` — fake S3 client; covers parse,
  read, cursor-resume, check, and discover. Uses `errors.Is` per the
  `noerrstringmatch` structural lint.
- `sources/<source>/testdata/*.json` — fixtures validated by the
  existing `tools/catalogcheck` contract check.
- `internal/sourceprojection/<source>.go` — projectors mapping each
  family to the graph. Image entities reuse the
  `gcp_artifact_registry_image` URN so this source correlates with
  existing runtime / registry sources without dedup work; findings
  reuse the canonical vulnerability entity.
- `internal/sourceprojection/registry.go` — register projectors.
- `internal/sourceregistry/registry.go` + `_test.go` — register the
  new source loader.
- `go.mod` / `go.sum` — `aws-sdk-go-v2/service/s3 v1.101.0` (peer of
  the version already used by `kolide` / `sentinelone`).

### Validation

Commands run:

- `go build ./...` — clean
- `go test ./...` — all 41 packages pass
- `tools/catalogcheck` — clean (every emitted kind has a projector;
  every fixture validates its contract)
- `make lint` (golangci-lint v2.11.4) — 0 issues
- `make check-structural` (cerebrolint) — clean
- `make check-arch` — clean

### Security

- [x] No secrets in diff
- [x] Secret scan: clean
- [x] New deps: `aws-sdk-go-v2/service/s3 v1.101.0` (peer of existing
      S3-backed sources)

### Deployment ordering

This PR is the first of three coordinated changes and must merge first
so the new source ships in the next Cerebro image build. The remaining
changes (stack wiring + producer-side exporter) live in internal repos
and will be sequenced behind this PR.

### Do NOT merge

Open for review only; merge will be coordinated with the companion
internal PRs.


sources/aurelius/source.go
```diff
@@ -0,0 +1,447 @@
...
+			}
+			events = append(events, event)
+			if rec.OccurredAt.After(watermark) {
+				watermark = rec.OccurredAt
+			}
+			if len(events) >= maxEventsPerPull {
+				break
+			}
+		}
+		lastKey = key
```
jonathanhaaswriter: [P2] Do not advance past partially processed archives

When a single NDJSON object contains more than `maxEventsPerPull` records, the inner loop breaks after emitting 10k events but this line still advances `lastKey` to the archive key. The next read uses `StartAfter=lastKey`, so S3 skips the same object entirely and every remaining record in that archive is permanently dropped.

sources/aurelius/source.go
```diff
@@ -0,0 +1,447 @@
...
+		if len(events) >= maxEventsPerPull {
+			break
+		}
+	}
+	pull := sourcecdk.Pull{Events: events}
+	if lastKey != "" {
```
jonathanhaaswriter: [P1] Stop returning the same cursor after the final page

This returns a non-nil `NextCursor` whenever `lastKey` is set, even when S3 returned the final page. After the caller reads with that cursor, an empty `ListObjectsV2` response leaves `lastKey` initialized to `startAfter` and returns the identical cursor again, so runtime sync never marks the source complete and repeatedly polls empty pages.

jonathanhaaswriter (changes requested):
Requesting changes due to S3 cursor handling issues that can cause repeated empty pulls or skipped records.

sources/aurelius/source.go
```diff
@@ -0,0 +1,447 @@
...
+	}
+	payload, err := json.Marshal(rec.Payload)
+	if err != nil {
+		return nil, fmt.Errorf("marshal payload: %w", err)
+	}
+	tenantID := strings.TrimSpace(rec.TenantID)
+	if tenantID == "" {
+		tenantID = st.tenantID
+	}
+	attributes := make(map[string]string, len(rec.Attributes))
```
jonathanhaaswriter, line 577: [P1] Keep finding CVE fields in attributes

`buildEvent` only copies `rec.Attributes` into the envelope, but the catalog/source tests allow `aurelius.finding` records where `cve_id` and `package` are payload-only. `aureliusFindingProjections` passes only `event.Attributes` to `addCanonicalVulnerabilityEntity`, so those accepted records silently project no vulnerability or `affected_by` edge. Please either promote the finding projector fields into attributes when emitting, or make them required attributes and update the fixtures/tests.

jonathanhaaswriter (approved):
The three previously requested Aurelius fixes are addressed: cursor state now resumes partially processed archives, final empty pages no longer loop the same cursor, and accepted CVE/package fields reach vulnerability projection attributes. Local and CI validation are passing.

PR #632, opened 2026-05-23
fix: avoid unsupported finding replays

jonathanhaaswriter:
## Summary
- skip append-log replay when selected finding rules are stale-only/unsupported for the runtime
- tighten event-rule runtime matching for explicit source families so unrelated runtimes do not trigger replay
- add coverage for explicit GRC family matching and no-replay stale cleanup paths

## Validation
- go test ./internal/findings -count=1
- make verify


jonathanhaaswriter:
Addressed the Droid review note in 9610c1da: event-rule runtime matching now uses the existing default-family-aware matcher, and tests cover GitHub default runtimes not matching github.audit rules while explicit audit and dependabot_alert runtimes still do. Local and CI make verify both pass.

PR #890, opened 2026-06-10
fix(source): allow private Panopticon API endpoints

jonathanhaaswriter:
## Summary
- add private_endpoint_allowlist support to the Panopticon source
- pass the allowlist into the hardened HTTP client for private DNS resolutions
- cover private endpoint allowlist parsing and resolved-private-IP behavior

## Validation
- go test ./sources/panopticon -count=1 -v
- go test ./sources/... -count=1
- go test ./... -count=1


jonathanhaaswriter:
Companion infra PR: WriterInternal/cerebro#1022 switches sec-dev/go-prod Panopticon runtimes to API mode after this source patch is available in a deployed image.

PR #909, opened 2026-06-11
feat: add OpenTelemetry support

jonathanhaaswriter:
## Summary

- Adds configurable OpenTelemetry OTLP tracing and metrics export for Cerebro.
- Bridges existing internal telemetry spans/events into OTEL while preserving stderr JSON telemetry.
- Instruments HTTP server requests with OTEL spans and metrics, and wires telemetry lifecycle into server and CLI execution paths.

## Test Plan

- `make verify`

## Known Invariants

- Source connectors use `internal/sourcehttp` for outbound HTTP safety.
- Production body reads are bounded with `io.LimitReader` or streamed.
- Graph Ask queries stay tenant-scoped, read-only, row-limited, and validated.
- Ask deterministic post-processing is not applied to LLM fallback rows.
- Candidate finding lifecycle writes remain atomic and idempotent.
- Public request origin, DPoP `htu`, and trusted proxy handling use the canonical origin helpers.

## Droid Review Context

- Fast local preflight: `make droid-review-preflight`.
- Focus review on changed behavior and the invariants above.
- Escalate to a deep manual Droid tag review for broad auth, graph, source HTTP, or state-machine changes.


jonathanhaaswriter:
Addressed the Droid review feedback in b979d7bb: error statuses now map to OTEL error status, source-runtime/orchestrator telemetry shutdown uses an uncanceled shutdown context, and inbound Traceparent is no longer trusted before auth while the sampler is ParentBased.

PR #1218, opened 2026-06-18
feat(tailscale): preserve device blocks-incoming retraction reason in projection telemetry

dcoln25-writer:
## Scope

Follow-up that extends the tailscale projection retraction reason allowlist so a bounded, tailscale-specific retraction reason is preserved in projection telemetry instead of collapsing to `unknown`. Source-scoped to tailscale projection and the shared bounded-reason helper.

## Source changes

None. The tailscale source already emits the device `blocks_incoming_connections` attribute consumed here.

## Projection changes

- Added `tailscale_device_blocks_incoming` to the `boundedProjectionRetractionReason` allowlist in `internal/sourceprojection/projector.go` so the reason survives telemetry bounding.
- A device that blocks incoming connections no longer projects an owner reachability edge, and any existing owner reachability edge is retracted with the bounded reason.
- Reused the existing device retraction code path; deauthorization retains precedence when both conditions are present.

## Finding changes

None. No finding rules, rulepack, audit, or detection catalog artifacts were touched.

## Shared file touches

- `internal/sourceprojection/projector.go` (bounded retraction reason allowlist only).

## Validation evidence

- `make lint` -> 0 issues.
- `go test ./sources/tailscale/... -count=1` -> ok.
- `go test ./internal/sourceprojection/... -count=1` -> ok (new tests matched and passed).
- `go test ./tools/archtests/... -count=1` -> ok.
- `make catalog-check` -> ok.
- `make detection-catalog-check` -> ok.

## Rollback notes

Revert this commit. The change is additive to the bounded-reason allowlist plus a single projection gating condition; no schema, rule, or catalog migrations are involved.

### DIFF
--- a/internal/sourceprojection/projector.go
+++ b/internal/sourceprojection/projector.go
@@ -195,6 +195,7 @@ func boundedProjectionRetractionReason(value string) string {
 		"cloudflare_dns_record_zone_reassigned",
 		"trivy_vulnerability_resolved",
 		"tailscale_device_deauthorized",
+		"tailscale_device_blocks_incoming",
 		"tailscale_grant_disabled":
 		return strings.TrimSpace(value)
 	default:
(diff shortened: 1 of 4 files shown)


jonathanhaaswriter (approved):
Checks are green and Droid found no issues; approving for merge.

PR #1287, opened 2026-06-19
fix: dedupe catalog validation issues

jonathanhaaswriter:
## Summary

- Dedupe catalog validation issues after aggregating checks.
- Add regression coverage to ensure policy DSL metadata issues appear once in repository validation output.

## Test Plan

- go test ./tools/catalogcheck -count=1 -v
- make catalog-check

## Known Invariants

- Source connectors use internal/sourcehttp for outbound HTTP safety.
- Production body reads are bounded with io.LimitReader or streamed.
- Graph Ask queries stay tenant-scoped, read-only, row-limited, and validated.
- Ask deterministic post-processing is not applied to LLM fallback rows.
- Candidate finding lifecycle writes remain atomic and idempotent.
- Public request origin, DPoP htu, and trusted proxy handling use the canonical origin helpers.

## Droid Review Context

- Fast local preflight: make droid-review-preflight.
- Focus review on changed behavior and the invariants above.
- Escalate to a deep manual Droid tag review for broad auth, graph, source HTTP, or state-machine changes.
- Land with make land-pr PR=<number> so Droid finishes before branch deletion.


jonathanhaaswriter:
Superseded by clean follow-up PR #1289 based on current main.

PR #1324, opened 2026-06-20
feat: add flag-gated per-runtime append-log replay index

jonathanhaaswriter:
## Summary

Adds an optional secondary index that accelerates runtime-scoped append-log replays.

- When enabled, replay resolves the bulk of a runtime's history from the index and **always merges the un-indexed tail** directly from the stream, so results stay correct and complete regardless of index freshness.
- When the flag is off or the index is unpopulated, replay **falls back** to the existing subject/legacy strategies with no behavior change.

## Changes

- **ports**: index entry/scan/query/result types and source/reader/writer interfaces.
- **state store**: additive Postgres tables for the index and its watermark, with parameterized newest-first lookups and optional exact-kind narrowing.
- **append log**: index-first replay strategy with mandatory tail-merge, plus a forward scan that emits per-runtime index entries.
- **population**: an incremental, idempotent maintenance job whose scan-and-persist loop lives in a dedicated domain package (`internal/appendlogindex`); bootstrap only wires dependencies, authorizes (admin), and adapts payloads.
- **config**: a single boolean flag, default off.

Disabled by default and gated end to end (schema, population, and read path) behind the flag.

## Tests

- New unit tests for the read strategy (index hit + tail-merge, fallback when unavailable, retention-purge skip, exact-kind tail), the forward scan, the Postgres store guards/schema/query builder, the population loop, the config flag, and job authorization.
- go build, go vet, gofmt, full go test, golangci-lint, and the architecture ratchet tests all pass.

## Compatibility

Additive and backward compatible. No schema or behavior changes occur unless the flag is explicitly enabled.


jonathanhaaswriter:
Addressed the P1 in b318394d. The index path is now gated by `runtimeIndexReplayEligible`: it serves a replay from the index only when the request is fully covered by the index's narrowing (runtime id, plus exact kinds in exact-kind mode). Non-exact kind-prefix, attribute-equals, and tenant-filtered replays fall back to the subject/legacy strategies, which collect matched candidates directly, so no matches below the watermark are dropped. Added regression tests for the non-exact kind-prefix case and for attribute/tenant filters skipping the index.

jonathanhaaswriter:
Addressed the P2 in b8008e81: documented `CEREBRO_JETSTREAM_RUNTIME_INDEX_ENABLED` in both `docs/reference/config-env-vars.md` and `docs/operations/hosting.md` alongside the other JetStream entries. `docs-drift-check` passes.

Issue #956, opened 2026-06-15
sources: shrink grandfathered connectors through Source CDK extraction


jonathanhaaswriter:
Follow-up from the 2026-06 Cerebro repo review and related to #684.

The source LOC ratchet is enforced, but several grandfathered source packages remain very large. The next improvement is to shrink them by extracting repeated transport, pagination, error mapping, and event-contract helpers into Source CDK/internal source helpers.

Acceptance criteria:
- Pick one large connector at a time and lower its exact grandfathered LOC ceiling when it shrinks.
- Do not raise grandfathered budgets.
- Prefer shared CDK/helper extraction over per-source special cases.
- Keep source package tests and catalog fixtures green.

jonathanhaaswriter:
Worked a small source shrink / CDK extraction slice locally:

- Added `sources/internal/textutil.FirstNonEmpty` and moved duplicated first-non-empty behavior in AWS, Azure, and Okta onto that shared helper.
- Added unit coverage for the helper.
- Ratcheted the grandfathered LOC budgets down for AWS, Azure, and Okta so the removed duplication cannot silently grow back.

This is intentionally a small slice of the larger extraction backlog, but it turns one repeated source helper into shared source infrastructure and tightens the architectural budget behind it.

Validation run locally:

- `go test ./sources/aws ./sources/azure ./sources/okta ./sources/internal/textutil -count=1`
- `make docs-drift-check readme-check catalog-check check-arch`
- `make test`

Issue #684, opened 2026-05-29
arch: enforce Source CDK size and I/O budgets


jonathanhaaswriter:
## Priority
High

## Problem
`NON_GOALS.md` and `PLAN.md` say sources should stay at or below 300 LOC and avoid direct I/O primitives. Current source implementations are much larger:

| Source | LOC |
| --- | ---: |
| `aws` | 2,568 |
| `okta` | 2,437 |
| `sentinelone` | 2,330 |
| `github` | 1,962 |

Some sources, such as `github`, import `net/http` directly instead of going through `sources/internal/jsonapi`. The budget is documented but not CI-enforced; `tools/archtests/source_packages_test.go` only checks catalog/testdata presence.

## Suggested direction
- Add an arch test with a grandfather list for existing sources and a hard cap for new sources.
- Add checks that new sources use the CDK/jsonapi boundary instead of direct I/O primitives.
- Migrate large sources incrementally, starting with a smaller pilot such as `github`.

## Acceptance criteria
- CI prevents new sources from exceeding the agreed LOC/I/O budget.
- Existing oversized sources are explicitly grandfathered with tracked migration work.
- At least one pilot source migration plan is identified.

jonathanhaaswriter:
Follow-up from the 2026-06 review pass: created #956 for shrinking grandfathered connectors via Source CDK/internal helper extraction. First local slice extracts repeated first-non-empty string logic into `sources/internal/textutil` and ratchets AWS/Azure/Okta package budgets down.

jonathanhaaswriter:
Follow-up from this pass: I worked one small source-budget ratchet locally.

- Added `sources/internal/textutil.FirstNonEmpty`.
- Moved duplicated first-non-empty helpers in AWS, Azure, and Okta onto the shared helper.
- Added unit coverage for the helper.
- Ratcheted the AWS, Azure, and Okta source package LOC budgets down by the removed lines.

Validation run locally:

- `go test ./sources/aws ./sources/azure ./sources/okta ./sources/internal/textutil -count=1`
- `make docs-drift-check readme-check catalog-check check-arch`
- `make test`

PR #1337, opened 2026-06-21
refactor(sources): extract shared host-safety into Source CDK (#956)

jonathanhaaswriter:
## What
Extracts the provider-agnostic host/IP safety classification out of the SentinelOne source connector into a new typed helper file in the Source CDK (`internal/sourcecdk/host_safety.go`), exposing `HostIsUnsafe` and `HostIsLoopback`. The connector now calls the shared helper instead of carrying its own copy.

The moved logic rejects loopback/private/link-local/unspecified/multicast targets and additionally normalizes obfuscated numeric IPv4 hosts (decimal/octal/hex forms) that `net.ParseIP` alone does not catch.

## Why
Part of the Source CDK extraction effort (#956, #684): shared request/safety plumbing should live in the CDK so legacy source packages shrink toward the LOC budget instead of duplicating common behavior across connectors.

## Notes
- Pure refactor: no change to emitted events, attributes, URNs, schema refs, cursors, or runtime validation behavior. The relocated code is identical logic, only moved and renamed at the two call sites.
- The helper uses typed signatures (string / net.IP / bool) to satisfy the CDK typed-boundary rule, and imports only `net` (not `net/http`), respecting the HTTP-client boundary.
- No-growth ratchet lowered for the affected source package (2181 to 2089) in both the archtest budget and the extraction-plan doc.

## Tests
- `go build ./...`
- `go test ./sources/sentinelone/... ./internal/sourcecdk/... -count=1`
- `make check-structural`

All green.

### DIFF
--- a/docs/engineering/source-cdk-extraction.md
+++ b/docs/engineering/source-cdk-extraction.md
@@ -32,5 +32,5 @@ Legacy sources above the 300 LOC budget are grandfathered with exact no-growth c
 | `grc` | 1378 | Keep control/evidence shaping out of source orchestration and push common mapping into shared helpers. |
 | `okta` | 2361 | Extract client, pagination, and identity/group/application readers into smaller units. |
 | `panopticon` | 820 | Separate request plumbing from emitted record construction. |
-| `sentinelone` | 2181 | Extract API paging, agent/application readers, and shared response normalization. |
+| `sentinelone` | 2089 | Extract API paging, agent/application readers, and shared response normalization. |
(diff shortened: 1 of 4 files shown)


jonathanhaaswriter:
Superseded by #1329, which introduces the shared internal/sourcecdk host-safety helper and migrates sentinelone onto it. sentinelone carried byte-identical SSRF classification logic to cosmo/okta, so consolidating avoids three PRs each re-adding the same helper file. Closing in favor of #1329.

PR #1349, opened 2026-06-21
refactor(sources): extract watermark string formatting into Source CDK (#956)

jonathanhaaswriter:
## Summary

Extract the shared watermark-string formatting used by the `aurelius` and `panopticon` sources into a new Source CDK helper, `sourcecdk.WatermarkString`, and delete the duplicated local `watermarkString` from both sources.

`WatermarkString(watermark, fallback time.Time) string` returns the later of the two times (preferring `watermark` on ties), rendered as a UTC RFC3339Nano string, and returns an empty string when both times are zero. Each source's call sites now call the helper directly.

> Stacked on #1345, #1346, and #1348. Base branch is `cdk-extract-checkpointcursor-20260621` so this PR shows only the watermark-formatting delta; it should merge after the PRs below it.

## Why

- Collapses an identical watermark-formatting helper that lived in two sources into the canonical Source CDK.
- Removes the corresponding entry from the cross-source duplication allowlist in `tools/archtests/source_helper_duplication_test.go` (and the matching backlog row), so the guardrail now enforces that this logic stays deduplicated.
- Ratchets the grandfathered per-source LOC budgets down to the new sizes (and updates the documented markers) to lock in the reduction.

## Behavior

No functional change. The helper is a behavior-preserving move of the previous logic, including UTC normalization.

## Tests

- `go build ./...`
- `go test ./internal/sourcecdk/... ./sources/aurelius/... ./sources/panopticon/... -count=1` (includes a new `WatermarkString` unit test covering later-of-two, tie, zero-input, both-zero, and UTC rendering cases)
- `go test ./tools/archtests/ -count=1` (duplication guardrail passes with the allowlist entry removed; LOC-budget ratchet and extraction-plan checks pass)
- `make check-structural`, `make docs-drift-check`, `make oss-audit`

Refs #956.

### DIFF
--- a/docs/engineering/source-cdk-extraction.md
+++ b/docs/engineering/source-cdk-extraction.md
@@ -22,7 +22,7 @@ Legacy sources above the 300 LOC budget are grandfathered with exact no-growth c
 
 | Source | Current LOC Budget | Extraction Pressure |
 | --- | ---: | --- |
-| `aurelius` | 623 | Split source orchestration from record mapping and shared request plumbing. |
+| `aurelius` | 612 | Split source orchestration from record mapping and shared request plumbing. |
 | `aws` | 19070 | Extract common AWS client/session, pagination, ARN parsing, and resource-family traversal helpers. |
 | `azure` | 2842 | Extract subscription traversal, client factories, and paginated resource readers. |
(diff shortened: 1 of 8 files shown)


jonathanhaaswriter:
Superseded by #1352, which lands this change (watermark string extraction) directly on main as part of a consolidated PR.

PR #1350, opened 2026-06-21
refactor(sources): extract JSON value formatting into Source CDK (#956)

jonathanhaaswriter:
## Summary

Extract the shared JSON value-to-string formatting used by the `cosmo` and `vulnview` sources into the Source CDK as a new `JSONScalar.Flattened()` method, and reduce each source's local `valueString` to a thin wrapper over it.

`Flattened()` renders a decoded JSON value (as produced by `encoding/json` into `map[string]any`): nil becomes empty, strings are trimmed, float64 values are formatted without an exponent, bools become `"true"`/`"false"`, `[]any` values become their comma-joined non-empty elements, and any other value is rendered via `fmt.Sprint` and trimmed. Scalar cases delegate to the existing `JSONScalar.String()`.

> Stacked on #1345, #1346, #1348, and #1349. Base branch is `cdk-extract-watermarkstring-20260621` so this PR shows only the JSON-value delta; it should merge after the PRs below it.

## Why

- Collapses an identical JSON value-formatting helper that lived in two sources into the canonical Source CDK, alongside the existing `JSONScalar` coercion helpers.
- Removes the corresponding entry from the cross-source duplication allowlist in `tools/archtests/source_helper_duplication_test.go` (and the matching backlog row), so the guardrail now enforces that this logic stays deduplicated. The remaining wrappers are below the guardrail threshold.
- Ratchets the grandfathered per-source LOC budgets down to the new sizes (and updates the documented markers) to lock in the reduction.

## Notes

The helper is exposed as a method on the existing `JSONScalar` box rather than a free function, to satisfy the Source CDK structural rule against exported functions with untyped (`any`) parameters.

## Behavior

No functional change; behavior-preserving move of the previous logic, including the recursive array handling.

## Tests

- `go build ./...`
- `go test ./internal/sourcecdk/... ./sources/cosmo/... ./sources/vulnview/... -count=1` (includes a new `JSONScalar.Flattened` unit test covering nil, trimmed string, float, bool, nested/empty-filtered slices, and the fmt.Sprint default)
- `go test ./tools/archtests/ -count=1` (duplication guardrail passes with the allowlist entry removed; LOC-budget ratchet and extraction-plan checks pass)
- `make check-structural`, `make docs-drift-check`, `make oss-audit`

Refs #956.

### DIFF
--- a/docs/engineering/source-cdk-extraction.md
+++ b/docs/engineering/source-cdk-extraction.md
@@ -25,7 +25,7 @@ Legacy sources above the 300 LOC budget are grandfathered with exact no-growth c
 | `aurelius` | 612 | Split source orchestration from record mapping and shared request plumbing. |
 | `aws` | 19070 | Extract common AWS client/session, pagination, ARN parsing, and resource-family traversal helpers. |
 | `azure` | 2842 | Extract subscription traversal, client factories, and paginated resource readers. |
-| `cosmo` | 1011 | Move shared API pagination and response normalization into reusable source helpers. |
+| `cosmo` | 992 | Move shared API pagination and response normalization into reusable source helpers. |
 | `gcp` | 2095 | Extract project traversal, service clients, and resource-family pagination helpers. |
 | `github` | 2028 | Split audit/event readers from repository/user normalization and shared pagination. |
 | `googleworkspace` | 815 | Extract API client setup and paginated directory readers. |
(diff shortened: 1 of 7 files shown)


jonathanhaaswriter:
Superseded by #1352, which lands this change (JSON value formatting extraction) directly on main as part of a consolidated PR.

PR #1478, opened 2026-06-26
Require indexed replay for finding rules

jonathanhaaswriter:
## Summary
- require runtime-indexed append-log replay for finding-rule evaluations
- fail fast when required runtime-index replay is unavailable instead of falling back to broad scans
- enable the JetStream runtime replay index by default, with an explicit env override to disable it

## Tests
- make -C /Users/jonathan/cerebro test
- make -C /Users/jonathan/cerebro lint

---

<a href="https://app.devin.ai/review/writer/cerebro/pull/1478" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/cmd/cerebro/orchestrator.go
+++ b/cmd/cerebro/orchestrator.go
@@ -353,7 +353,7 @@ func runOrchestratorLoop(ctx context.Context, options orchestratorOptions) (resu
 		findingEvaluationRunStore(deps.StateStore),
 		findingEvidenceStore(deps.StateStore),
 		claimStore(deps.StateStore),
-	).WithGraphStore(sourceProjectionGraphStore(deps.GraphStore)).WithGraphQueryStore(findingGraphQueryStore(deps.GraphStore)).WithAppendLog(deps.AppendLog).WithGraphRuleQueryTimeout(graphRuleQueryBudgetForPhase(options.PhaseTimeout))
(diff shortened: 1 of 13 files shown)


jonathanhaaswriter:
Addressed the CI failure in 31db033b by replacing err.Error() string matching with a typed sentinel checked via errors.Is; local structural checks now pass.

On the Devin note: the fail-fast behavior is intentional for finding-rule replays. The previous fallback could pin workers on broad stream scans and recreate the 15 minute timeout; requiring the runtime index makes the dependency explicit and fails fast until the index is ready. The index is now enabled by default, can still be explicitly disabled, and the failure is surfaced through telemetry rather than silently falling back.

jonathanhaaswriter:
Addressed the additional runtime-index population-lag feedback in 927c291e. Finding-rule replay paths now run a bounded runtime-index warmup before replaying; if the index is still behind after the bounded pass, evaluation fails fast with an explicit warming error and still does not fall back to broad stream scans. Wired this through App and standalone orchestrator paths, with regression coverage for warmup success/failure.

jonathanhaaswriter:
Addressed the missed review feedback in d378f74a. The Connect/RPC bootstrap finding service factories now wire the same runtime-index replay preparer as the App/orchestrator paths, with regression coverage across core, candidate, and workflow service construction. I also increased the bounded replay-preparation default budget so cold-start index catch-up gets a larger bounded pass before returning the warming error.

jonathanhaaswriter:
Follow-up correction pushed in 81fd52ea. `RequireRuntimeIndex` is now tied to the runtime-index preparer being enabled, so explicit runtime-index disablement no longer sends required-index replay requests. Added enforcing test coverage to catch mismatches between required-index requests and index readiness, including the disabled-index path.

PR #1555, opened 2026-06-29
Harden JetStream append-log routing

jonathanhaaswriter:
Summary
- Add CEREBRO_JETSTREAM_STREAM_NAME and publish with the expected-stream header for appends, canary checks, and replay readiness.
- Add an optional append batch capability for high-volume ingest paths without changing the base append-log interface.
- Keep replay results newest-observed-first by using occurred_at in the runtime index and by preserving newer event-time candidates during legacy scans.
- Filter reserved NATS headers out of workflow event attributes and update docs, examples, and deploy helpers.

Tests
- go test ./... -count=1
- python3 -m unittest scripts.tests.test_agent_onboard

---

<a href="https://app.devin.ai/review/writer/cerebro/pull/1555" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


jonathanhaaswriter:
Devin called out that legacy reverse replay now scans to the stream start instead of stopping after `candidateLimit` matches. That is intentional for this fallback path: stopping early can return the wrong result once replay ordering is newest-observed-first by `occurred_at`, because an older stream sequence can carry a newer observed timestamp.

The common runtime replay path is the per-runtime index, now documented as enabled by default; exact-kind replays also use the subject-index path. The full reverse scan remains the correctness fallback for filters the index cannot safely pre-apply, and the test added in this PR covers the missed-newer-observed-event case.

jonathanhaaswriter:
Ready for human review. All required checks are green; GitHub branch protection is currently blocking merge only on required review.

Internal deploy wiring has already merged and deployed through WriterInternal/cerebro#1919, so the runtime PR is the remaining step before promoting a runtime image with the app-side expected-stream behavior.

PR #1575, opened 2026-06-29
[codex] Add evidence-backed questionnaire answers

jonathanhaaswriter:
## Summary
- Add a graph-backed `questionnaire_evidence_answer` intent for QAuto-style compliance questions.
- Route Okta access/MFA/lifecycle, policy-document, and control-coverage prompts through bounded graph retrieval before LLM summarization.
- Extend questionnaire answer records with a small reasoning contract covering intent, relevant controls, source and policy citations, freshness, confidence, unsupported claims, and manual-review state.
- Preserve partial and blocked states so missing or stale evidence is visible instead of promoted as a complete answer.

## Validation
- `go test ./internal/graphagent ./internal/evidencepackets ./internal/bootstrap -count=1`
- `make check-structural`
- `make openapi-ts-check`
- `make contracts-check`
- `make test`
- `git diff | rg -n "(?i)vanta|comparison|inspiration"` returned no matches


internal/graphagent/query_plan.go
```diff
@@ -506,6 +566,84 @@ RETURN source.urn AS source_urn,
...
+       exception.urn AS exception_urn,
+       coalesce(exception.label, exception.urn) AS exception_label,
+       exception.entity_type AS exception_type,
+       exceptionRel.relation AS exception_relation,
+       coalesce(exception.attributes_json, '') AS exception_attributes_json_internal,
+       source.urn AS source_urn,
+       coalesce(source.label, source.urn) AS source_label,
+       coalesce(source.attributes_json, '') AS source_attributes_json_internal
+ORDER BY control_label, support_label, evidence_label
+LIMIT %d`, cypherJSONStringAttributes("control.attributes_json", "policy_type"), cypherJSONStringAttributes("control.attributes_json", "control_external_id", "control_id", "policy_id"), topicPredicate, limit), true
```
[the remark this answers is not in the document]
jonathanhaaswriter: Handled in `f7889f74a`: questionnaire evidence now uses a dedicated `questionnaireEvidenceCandidateRowLimit` of 500 instead of the global `ports.MaxCypherQueryRows` candidate window. Validation and saturation checks use the per-intent candidate limit, and the user-facing rows are still deduped and trimmed to the plan limit. Verified with `go test ./internal/graphagent`.

internal/graphagent/query_plan.go
```diff
@@ -250,6 +253,59 @@ func fastPathTopRiskFilters(question string) map[string]string {
 	return filters
 }
 
+func looksLikeQuestionnaireEvidenceQuestion(haystack string) bool {
+	trimmed := strings.TrimSpace(haystack)
+	if strings.Contains(haystack, "qauto") || strings.Contains(haystack, "questionnaire") || strings.Contains(haystack, "security questionnaire") {
+		return true
+	}
+	if strings.Contains(haystack, "evidence packet") {
+		return true
```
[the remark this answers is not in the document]
jonathanhaaswriter: Handled in `f7889f74a`: `evidence packet` no longer routes to questionnaire evidence by itself. The classifier now requires questionnaire/control/compliance context for that phrase, and generic questions like “What is in the evidence packet for this finding?” and “Show the evidence packet export status” stay on raw Cypher planning. Verified with `go test ./internal/graphagent`.

internal/evidencepackets/packets.go
```diff
@@ -20,30 +20,29 @@ import (
 const ContractVersion = "2026-06-24"
```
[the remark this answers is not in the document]
jonathanhaaswriter: Addressed in c5cf60947 by bumping evidencepackets.ContractVersion to 2026-06-29 for the questionnaire answer response fields. Verified with `go test ./internal/evidencepackets ./internal/bootstrap`.

internal/graphagent/query_plan.go
```diff
@@ -250,6 +254,79 @@ func fastPathTopRiskFilters(question string) map[string]string {
...
+		(strings.Contains(haystack, "answer") ||
+			strings.Contains(haystack, "audit") ||
+			strings.Contains(haystack, "compliance") ||
+			strings.Contains(haystack, "question") ||
+			strings.Contains(haystack, "show")) {
+		return true
+	}
+	if strings.Contains(haystack, "policy doc") || strings.Contains(haystack, "policy document") {
+		return true
+	}
```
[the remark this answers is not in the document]
jonathanhaaswriter: Addressed in 1da20f155 by requiring policy-document questionnaire context before routing to questionnaire evidence. Added regression coverage for non-questionnaire policy document inventory/count questions. Verified with `go test ./internal/graphagent`.

internal/graphagent/ask.go
```diff
@@ -385,13 +386,34 @@ func cypherRowsToMaps(rows []ports.CypherRow) []map[string]any {
...
+			mergeInternalAttributesWithPrefix(row, "finding_attributes_json_internal", "finding_", []string{"summary", "status", "severity", "effective_severity", "risk_score"})
+			mergeInternalAttributesWithPrefix(row, "exception_attributes_json_internal", "exception_", []string{"status", "exception_id", "expires_at", "owner_id", "source_system"})
+			mergeInternalAttributesWithPrefix(row, "source_attributes_json_internal", "source_", []string{"status", "health", "last_sync_at", "last_sync_minutes", "last_success_at", "last_error"})
+			mergeInternalAttributes(row, "relation_attributes_json_internal", []string{"severity", "effective_severity", "risk_score"})
+			removeRawAttributeJSONFields(row)
+			continue
+		}
 		mergeInternalAttributes(row, "finding_attributes_json_internal", []string{"summary", "status", "severity", "effective_severity", "risk_score"})
+		mergeInternalAttributes(row, "exception_attributes_json_internal", []string{"status", "exception_id", "expires_at", "owner_id", "source_system"})
```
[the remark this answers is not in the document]
jonathanhaaswriter: Addressed by the existing exception-prefix change: `exception_attributes_json_internal` is merged with the `exception_` prefix on both questionnaire and non-questionnaire sanitization paths, so fields like `status` no longer collide with finding/source status. Verified with `go test ./internal/graphagent`.

internal/graphagent/query_plan.go
```diff
@@ -506,6 +592,84 @@ RETURN source.urn AS source_urn,
...
+        WHEN control.urn = $scope_urn THEN true
+        ELSE false
+      END
+OPTIONAL MATCH (support)-[supportEvidenceRel:RELATION {relation: 'has_evidence'}]->(supportEvidence:Entity {tenant_id: $tenant_id})
+WITH DISTINCT control, control_ref, support, supportRel, supportEvidence, supportEvidenceRel
+OPTIONAL MATCH (control)-[controlEvidenceRel:RELATION {relation: 'has_evidence'}]->(controlEvidence:Entity {tenant_id: $tenant_id})
+WHERE supportEvidence IS NULL
+WITH DISTINCT control, control_ref, support, supportRel,
+     coalesce(supportEvidence, controlEvidence) AS evidence,
+     coalesce(supportEvidenceRel, controlEvidenceRel) AS evidenceRel
```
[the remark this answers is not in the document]
jonathanhaaswriter: Addressed in `67cada120`: the questionnaire template no longer suppresses direct control evidence behind `WHERE supportEvidence IS NULL`. It now carries direct control evidence as `direct_evidence_*` fields while still preserving the support-path evidence fields, and the regression test asserts both the direct fields and absence of the suppressing predicate. Verified with `go test ./internal/graphagent`.

internal/graphagent/query_plan.go
```diff
@@ -571,6 +735,41 @@ func resourceTypePredicate(value string) string {
...
+        WHEN qauto_match_text CONTAINS 'sso' THEN true
+        WHEN qauto_match_text CONTAINS 'group' THEN true
+        ELSE false
+      END`
+	case "policy_documents":
+		return "WHERE qauto_match_text CONTAINS 'policy' OR qauto_match_text CONTAINS 'document'"
+	case "control_coverage":
+		return "WHERE qauto_match_text CONTAINS 'control' OR qauto_match_text CONTAINS 'coverage' OR qauto_match_text CONTAINS 'evidence'"
+	default:
+		return ""
```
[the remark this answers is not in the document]
jonathanhaaswriter: Addressed in `e40bc0680`: `qauto_match_text` is built from labels, source IDs, control refs, and extracted attribute values rather than raw JSON blobs, so JSON key names like `policy_type` no longer make every row match. The `policy_documents` predicate now requires document-specific tokens, and `control_coverage` no longer uses `CONTAINS control`. Added regression coverage for both predicates. Verified with `go test ./internal/graphagent`.

internal/graphagent/ask.go
```diff
@@ -385,13 +386,34 @@ func cypherRowsToMaps(rows []ports.CypherRow) []map[string]any {
...
+			mergeInternalAttributesWithPrefix(row, "support_attributes_json_internal", "support_", []string{"evidence_type", "document_type", "policy_document_type", "policy_type", "questionnaire_type", "status", "source_system"})
+			mergeInternalAttributesWithPrefix(row, "evidence_attributes_json_internal", "evidence_", []string{"evidence_id", "evidence_type", "document_type", "policy_document_type", "status", "source_system"})
+			mergeInternalAttributesWithPrefix(row, "finding_attributes_json_internal", "finding_", []string{"summary", "status", "severity", "effective_severity", "risk_score"})
+			mergeInternalAttributesWithPrefix(row, "exception_attributes_json_internal", "exception_", []string{"status", "exception_id", "expires_at", "owner_id", "source_system"})
+			mergeInternalAttributesWithPrefix(row, "source_attributes_json_internal", "source_", []string{"status", "health", "last_sync_at", "last_sync_minutes", "last_success_at", "last_error"})
+			mergeInternalAttributes(row, "relation_attributes_json_internal", []string{"severity", "effective_severity", "risk_score"})
+			removeRawAttributeJSONFields(row)
```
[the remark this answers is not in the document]
jonathanhaaswriter: Addressed in `2044e6d743`: questionnaire row sanitization now merges `direct_evidence_attributes_json_internal` with the `direct_evidence_` prefix before raw internal JSON fields are removed. The regression test covers `direct_evidence_status`, `direct_evidence_evidence_type`, `direct_evidence_source_system`, and verifies the raw direct evidence attributes do not leak. CI has restarted for that commit.

internal/graphagent/query_plan.go
```diff
@@ -250,6 +254,84 @@ func fastPathTopRiskFilters(question string) map[string]string {
...
+		"lifecycle",
+	} {
+		if strings.Contains(haystack, phrase) {
+			return true
+		}
+	}
+	return false
+}
+
+func fastPathQuestionnaireEvidenceFilters(question string) map[string]string {
```
haasonsaas: **Questionnaire questions with no matched topic retrieve the entire control catalog.**

`looksLikeQuestionnaireEvidenceQuestion` routes a question to `IntentQuestionnaireEvidence` on the bare word "questionnaire"/"qauto", but this function only assigns a `topic` for the 5 hardcoded cases. For anything else (e.g. *"Answer this security questionnaire item"*, *"Is data encrypted at rest in S3?"*) `filters` has no `topic`, so `questionnaireEvidenceTopicPredicate` returns `""` and the emitted Cypher has **no `qauto_match_text` filter** — it returns every control + evidence (capped at 500) and feeds that unrelated set to the LLM summary. The deterministic fast path that was meant to *bound* retrieval degrades to "all controls in the tenant."

Consider returning `false` (fall back to LLM planning) when no topic resolves, rather than emitting an unfiltered template.

internal/graphagent/query_plan.go
```diff
@@ -571,6 +755,41 @@ func resourceTypePredicate(value string) string {
...
+		return `WHERE CASE
+        WHEN NOT qauto_match_text CONTAINS 'okta' THEN false
+        WHEN qauto_match_text CONTAINS 'access' THEN true
+        WHEN qauto_match_text CONTAINS 'sso' THEN true
+        WHEN qauto_match_text CONTAINS 'group' THEN true
+        ELSE false
+      END`
+	case "policy_documents":
+		return "WHERE qauto_match_text CONTAINS 'policy document' OR qauto_match_text CONTAINS 'policy_document' OR qauto_match_text CONTAINS 'document'"
+	case "control_coverage":
```
haasonsaas, line 780: **This `control_coverage` predicate is a near no-op.**

`qauto_match_text` is a concatenation that always includes `coalesce(evidence.entity_type,'')`, and evidence nodes have `entity_type = 'runtime_evidence'` — which contains the substring `'evidence'`. Since the query's `WHERE evidence IS NOT NULL OR support IS NOT NULL` guarantees evidence is usually present, `CONTAINS 'evidence'` matches essentially every candidate row, so the coverage/gap scoping is illusory.

Match on a word-boundaried or attribute-derived signal instead of a substring that collides with the baked-in entity type.

internal/evidencepackets/packets.go
```diff
@@ -1106,6 +1212,407 @@ func packetsForExpectation(control ControlPosture, request EvidenceRequest, item
...
+	if request.Required && len(packets) == 0 {
+		return QuestionnaireAnswerConfidence{Level: "low", Score: 10, Reason: "Required evidence has not been collected."}
+	}
+	if len(gaps) != 0 {
+		return QuestionnaireAnswerConfidence{Level: "low", Score: 25, Reason: "Evidence gaps prevent a supported answer."}
+	}
+	return QuestionnaireAnswerConfidence{Level: "low", Score: 20, Reason: "No supported answer is available from the current packet."}
+}
+
+func answerFreshness(request EvidenceRequest, packets []EvidencePacket) EvidenceFreshness {
```
haasonsaas, line 1489: **Freshness times are compared as lexicographic strings, which is wrong across timezone offsets.**

This aggregates `packet.Freshness.ObservedAt/ExpiresAt` via `maxTimeString`/`minTimeString` (raw string compare). Those strings come from `formatTime` = `value.Format(time.RFC3339)` with **no `.UTC()`**. If any upstream time carries a non-UTC offset, lexical order diverges from chronological order: `"2026-01-01T15:00:00Z"` (15:00Z) sorts *before* `"2026-01-01T23:00:00+09:00"` (14:00Z), so the reported nearest `expires_at` is wrong and overstates remaining freshness.

Latent today (proto times are UTC) but `minTimeString` now feeds an auditor-facing `expires_at`. Normalize to UTC or compare parsed `time.Time`.

internal/graphagent/query_plan.go
```diff
@@ -506,6 +592,104 @@ RETURN source.urn AS source_urn,
        coalesce(source.attributes_json, '') AS source_attributes_json_internal
 ORDER BY source_label, source_urn
 LIMIT %d`, limit), true
+	case IntentQuestionnaireEvidence:
```
haasonsaas, line 572: **The 500-row candidate cap is applied before dedup, so high-fanout tenants silently lose later controls.**

This template `LIMIT`s at `questionnaireEvidenceCandidateRowLimit` (500) *before* `postProcessQuestionnaireEvidenceRows` dedups by URN tuple, and rows are `ORDER BY control_label`. A few controls with many support/evidence/finding edges can consume all 500 rows, so controls sorted later never enter the candidate set and produce no answer — with no truncation signal to the caller (the other intents use the 3000 `MaxCypherQueryRows` ceiling).

Consider surfacing a "candidates truncated" flag.

internal/evidencepackets/packets.go
```diff
@@ -395,6 +402,96 @@ type EvidenceExportArtifact struct {
 	Included    bool   `json:"included"`
 }
 
+type EvidenceExportRecords struct {
```
haasonsaas: **`EvidenceExportRecords` is dead code.**

This struct is never referenced anywhere — it looks like a half-finished mirror of the `ResponseQuestionnaireRecords` embedding refactor that was never wired into `Response` (which still declares `Artifacts`/`Export`/`Snapshot` as inline fields). It also carries the same `json` tags, so a future "finish the refactor" that embeds it while leaving the inline fields would create duplicate promoted fields.

Delete it or complete the refactor.

internal/evidencepackets/packets.go
```diff
@@ -1106,6 +1212,407 @@ func packetsForExpectation(control ControlPosture, request EvidenceRequest, item
...
+		if _, ok := seen[value]; ok {
+			continue
+		}
+		seen[value] = struct{}{}
+		unique = append(unique, value)
+	}
+	return unique
+}
+
+func answerText(control ControlPosture, request EvidenceRequest, state string) string {
```
haasonsaas, line 1345: **The "answer" is a canned status label, not derived from evidence.**

`answerText` returns fixed English templates keyed only on state, and `answerConfidence` hardcodes magic scores (90/60/50/25/20/10) with no central contract. For a feature named "evidence-backed answers," `questionnaire_answers[].answer` never incorporates evidence content and isn't localizable.

Reasonable as a v1 placeholder — flag the scores in particular, since they'll drift from any UI threshold logic that consumes them.

internal/graphagent/query_plan.go
```diff
@@ -506,6 +592,104 @@ RETURN source.urn AS source_urn,
...
+       coalesce(exception.attributes_json, '') AS exception_attributes_json_internal,
+       source.urn AS source_urn,
+       coalesce(source.label, source.urn) AS source_label,
+       coalesce(source.attributes_json, '') AS source_attributes_json_internal
+ORDER BY control_label, support_label, evidence_label
+LIMIT %d`, cypherJSONStringAttributes("control.attributes_json", "policy_type"), cypherJSONStringAttributes("control.attributes_json", "control_external_id", "control_id", "policy_id"), cypherJSONStringAttributes("control.attributes_json", "control_external_id", "control_id", "policy_id"), cypherJSONStringAttributes("support.attributes_json", "evidence_type", "document_type", "policy_document_type", "policy_type", "questionnaire_type", "source_system"), cypherJSONStringAttributes("evidence.attributes_json", "evidence_type", "document_type", "policy_document_type", "policy_type", "questionnaire_type", "source_system"), cypherJSONStringAttributes("controlEvidence.attributes_json", "evidence_type", "document_type", "policy_document_type", "policy_type", "questionnaire_type", "source_system"), topicPredicate, questionnaireEvidenceCandidateRowLimit), true
```
haasonsaas: **~100-line Cypher assembled via 8 positional `fmt.Sprintf` args is fragile.**

This splices six `cypherJSONStringAttributes(...)` results, a raw `%s` `topicPredicate` mid-query, and a `%d` limit purely by position. The team already needed exact-substring assertion tests to guard it; a future edit that adds/reorders a `coalesce(...)` column will mis-bind an attribute extraction to the wrong node with no compile error (e.g. support attributes rendered into the control column).

Consider named placeholders or composing the query from smaller fragments.

internal/graphagent/query_plan.go
```diff
@@ -571,6 +755,41 @@ func resourceTypePredicate(value string) string {
...
+      END`
+	case "okta_lifecycle":
+		return `WHERE CASE
+        WHEN NOT qauto_match_text CONTAINS 'okta' THEN false
+        WHEN qauto_match_text CONTAINS 'lifecycle' THEN true
+        WHEN qauto_match_text CONTAINS 'deprovision' THEN true
+        WHEN qauto_match_text CONTAINS 'provision' THEN true
+        ELSE false
+      END`
+	case "okta_access":
```
haasonsaas, line 770: **Substring `CONTAINS` here has false positives.**

`CONTAINS 'access'` matches `accessory`/`accessible`; `CONTAINS 'group'` matches any group reference. Gated by `'okta'` so impact is bounded, but it can attribute an unrelated control's evidence to an access answer. Same word-boundary concern as the `control_coverage` predicate below.

internal/evidencepackets/packets.go
```diff
@@ -1106,6 +1212,407 @@ func packetsForExpectation(control ControlPosture, request EvidenceRequest, item
...
+
+func missingEvidenceIDs(gaps []QuestionnaireEvidenceGap) []string {
+	ids := make([]string, 0, len(gaps))
+	for _, gap := range gaps {
+		ids = append(ids, gap.ID)
+	}
+	return uniqueStrings(ids)
+}
+
+func uniqueStrings(values []string) []string {
```
haasonsaas, line 1328: **`uniqueStrings` duplicates `uniqueSortedStrings`.**

This is byte-identical trim/skip-empty/dedup logic to `uniqueSortedStrings` (further down this file), differing only by the trailing `sort.Strings`. Have one call the other so the trimming/dedup behavior can't silently diverge.

haasonsaas (commented):
Automated review (recall-biased, high effort). 10 findings: 4 correctness/behavior, 6 cleanup/altitude. Build + package tests pass. Posted inline below, ranked most-severe first.

internal/graphagent/query_plan.go
```diff
@@ -250,6 +254,84 @@ func fastPathTopRiskFilters(question string) map[string]string {
 	return filters
 }
 
+func looksLikeQuestionnaireEvidenceQuestion(haystack string) bool {
```
haasonsaas, line 260: **Keyword-soup intent classifier is brittle and order-dependent.**

This is a large OR-tree whose correctness depends on `inferIntent` checking identity-bridge/explain-finding *first* (that's the only reason *"Which identities bridge Okta and GitHub access?"* doesn't misroute). A phrasing like *"audit answer for access posture"* hits the generic `access`+`audit` keywords, and it can't generalize to new frameworks without editing the list.

Related cleanup: the answer-context cluster (`"answer"||"audit"||"compliance"||"question"`) is copy-pasted across three branches — extract a helper.

jonathanhaaswriter:
Closing this in favor of the replacement PR: #1591.

PR #1583, opened 2026-06-30
Deepen sourcegen fixture coverage

jonathanhaaswriter:
## Summary

- Generate discover and read fixture files for every Source CDK family.
- Add runtime-depth counters for read fixtures, discover fixtures, and complete fixture pairs.
- Include the fixture-pair contract in generated runtime docs and PR notes.

## Harness result

- Runtime-backed sources: 794
- Reference-runtime sources: 8
- Sources with read fixtures: 787
- Sources with discover fixtures: 8
- Sources with complete fixture pairs: 8

## Tests

- go test ./internal/sourcegen ./internal/connectorcatalog ./tools/connectorcatalogreview ./tools/connectorcatalogfidelity
- make connector-catalog-maintenance


jonathanhaaswriter:
Moving this back to draft while #1588 lands. This fixture-generation improvement can still be useful, but it should not be treated as runtime-depth proof by itself; it needs to rebase behind the provider API contract gate and keep the new provider-contract counters intact.

PR #1825, opened 2026-07-14
Share Wasm guest ABI mechanics

jonathanhaaswriter:
## What changed

- Add internal/wasmguest for allocation, Wasm32 range checks, input slices, fixed response writes, and the shared 16-byte JSON descriptor.
- Migrate the static validator, MITRE evaluator, source coverage evaluator, and Panopticon resource extractor to the shared mechanics.
- Route shared guest-crate changes through the aggregate embedded Wasm check.
- Add raw ABI integration coverage that invokes all four checked-in artifacts at null, zero-length, overlapping, and out-of-bounds memory boundaries.

## Compatibility

- Export names, function signatures, ABI versions, and domain status codes are unchanged.
- The static validator retains its 24-byte decision protocol.
- JSON guests retain their 16-byte status, pointer, length, and reserved descriptor.

## Validation

- Canonical Linux x86_64 artifacts rebuilt for all four guests
- Exact changed-check plan against #1824
- Rust workspace format, clippy, test, and wasm32 release builds
- Raw ABI normal and race tests for all four artifacts
- Focused Go host normal and race tests
- Architecture, structural, scoped lint, and diff checks

## Stack

Depends on #1824. Review only the shared guest ABI commit in this PR.


jonathanhaaswriter:
Superseded by draft #1852. Exact cumulative accounting from the #1825 base through #1827 is 42 files and 2,341 changed lines, with path hash `9ac5a28fecb650a8914084c119c0b2d247905d3c409597528a9c28a0902aac00`; #1852 preserves that full path set and accounting. All 42 paths remain present. Forty blobs are byte-identical; only the validator source and its regenerated Wasm artifact are intentionally updated to carry the foundation security fixes. This draft carries no remaining unique scope and is closed without merge.

PR #1826, opened 2026-07-14
Share embedded Wasm host lifecycle

jonathanhaaswriter:
## What changed

- Add internal/wasmhost for lazy compilation, no-import and export-signature validation, ABI checks, memory limits, initialization bounds, per-call timeouts, fresh instances, and Wasm32 pointer checks.
- Migrate the JSON evaluator host and static validator host to the shared lifecycle.
- Preserve cached configuration and initialization failures, including canceled-first-call recovery.
- Keep each domain package responsible for its result decoder and status mapping.

## Compatibility

- Guest artifacts and guest export contracts are unchanged.
- The JSON 16-byte descriptor remains owned by internal/wasmjson.
- The static validator 24-byte decision protocol, safety policy, limits, and fail-closed behavior are unchanged.

## Validation

- Exact changed-check plan against #1825
- Focused Go tests and race tests across the shared host and all consumers
- Repeated initialization, cancellation, concurrency, ABI, and pointer tests
- Embedded static-validator artifact check
- Architecture, structural, scoped lint, and diff checks

## Stack

Depends on #1825. Review only the shared host lifecycle commit in this PR.


jonathanhaaswriter:
Superseded by draft #1852. Exact cumulative accounting from the #1825 base through #1827 is 42 files and 2,341 changed lines, with path hash `9ac5a28fecb650a8914084c119c0b2d247905d3c409597528a9c28a0902aac00`; #1852 preserves that full path set and accounting. All 42 paths remain present. Forty blobs are byte-identical; only the validator source and its regenerated Wasm artifact are intentionally updated to carry the foundation security fixes. This draft carries no remaining unique scope and is closed without merge.

PR #1859, opened 2026-07-15
Start HTTP before durable projection recovery

jonathanhaaswriter:
## What changed
- start durable projection and platform-job recovery in the existing shutdown-tracked background supervisor
- let the HTTP server bind before a large assessment replay completes
- retry transient recovery failures before continuous job and assessment reconciliation starts
- cover blocked replay, retry, and cancellation behavior with regression tests

## Why
Production tasks were terminated by load-balancer health checks while startup synchronously scanned more than 1.5 million assessment events. The API never opened its health port, so the previous task revision remained active. A transient replay failure also must not leave an API serving indefinitely with incomplete recovered projections.

## Verification
- go test ./internal/bootstrap ./cmd/cerebro
- git diff --check origin/main...HEAD

### DIFF
--- a/cmd/cerebro/main.go
+++ b/cmd/cerebro/main.go
@@ -141,9 +141,6 @@ func serve() error {
 	}
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer stop()
-	if _, err := app.RecoverPlatformJobs(ctx); err != nil {
-		return fmt.Errorf("recover platform jobs: %w", err)
-	}
 	jobRecoveryDone := app.StartPlatformJobRecovery(ctx, log.Printf)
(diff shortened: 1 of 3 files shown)


jonathanhaaswriter:
## Candidate canary receipt

- Candidate Build: https://github.com/writer/cerebro/actions/runs/29400482380
- Commit: 0c1918fb5b0fdc58202773452eb3ad0f30d91e9e
- Image digest: sha256:b656156464981e17fa20ad9f0b873ff056669dda28c08b9dfa15c9c63f545b76
- Result: an isolated ARM64 task using the current production configuration reached ECS task health HEALTHY and API container health HEALTHY.
- Started: 2026-07-15T08:42:19Z
- Stopped: 2026-07-15T08:43:42Z
- Cleanup: the canary task stopped and the temporary task definition was deregistered. The live service was not updated.

PR #1860, opened 2026-07-15
Harden graph action generator contracts

jonathanhaaswriter:
## Summary

- reject conflicting generator modes before filesystem work and return successful help
- reject unknown catalog fields and preserve error categories with structured Rust errors
- make the supported file-generation platform contract explicit and cover it with focused tests

## Testing

- `make graph-action-check`


jonathanhaaswriter:
Landing hold: keep this PR draft while the assessment response-contract follow-up and validator security fix establish the next main commit. Re-pin and re-gate after that main commit; do not mark ready or merge during this window.

PR #1936, opened 2026-07-16
build: add monorepo workspace scaffold

jonathanhaaswriter:
## Summary

- add root npm workspaces for applications and packages
- add shared TypeScript configuration and workspace validation commands
- retain platform-specific TypeScript packages required by CI
- add architecture and leak checks for the public monorepo boundary

This change contains repository mechanics only. It does not add deployment configuration, infrastructure resources, credentials, or environment topology.

## Validation

- `npm run check:workspaces`
- `make check-structural`
- `bash scripts/leak_check.sh range origin/main...HEAD`
- `git diff --check origin/main...HEAD`

## Sequence

This PR can merge directly into `main`. Application imports can then target this scaffold without depending on separate runtime contract work.


jonathanhaaswriter:
The `structural` failure is shared baseline drift from the current `main`, with `verify` failing transitively. #1940 removes the forbidden background context and now also passes an explicit `nobackgroundctx` run over `./tools/testauthor`, which the normal structural package set does not cover. I will refresh this branch and rerun its checks after #1940 lands; no scaffold source change is required for that failure.

jonathanhaaswriter:
Current head `f0edccdd8c2994bde28c7e58788213dff230217d` is current with `main` and all hosted checks pass. Workspace SDK tests, architecture, structural, and disclosure gates also pass locally; there are no unresolved review threads. Ready for repository review.

PR #1938, opened 2026-07-16
feat(slack): add durable transport boundary

jonathanhaaswriter:
## Summary

- verify Events API requests over exact request bytes with bounded timestamp skew
- persist the authenticated inbound payload before normalization, admission, or acknowledgement
- add Socket Mode presence proofs and acknowledgement ordering
- expose topology-neutral ingress readiness requirements
- keep URL verification isolated from event admission

The package defines portable transport behavior only. It contains no ingress resources, signing values, network routes, or environment configuration.

## Correctness properties

- invalid, stale, or unauthenticated Events API requests are not acknowledged
- a Slack event is acknowledged only after its payload and admitted run are durable
- inactive Socket Mode connections cannot acknowledge envelopes
- Events API readiness requires public ingress capability; Socket Mode readiness requires a durable presence relay

## Validation

- Slack companion workspace check: 18 tests
- focused transport suite: 10 tests
- OSS audit and public-boundary leak scan
- Practice Registry final gate
- `git diff --check`

### DIFF
--- a/apps/slack-companion/src/transport/contracts.ts
+++ b/apps/slack-companion/src/transport/contracts.ts
@@ -1,0 +1,161 @@
+import type { CapabilityRequirement, RunKind } from "@writer/cerebro-sdk";
+import type {
+  SlackAdmissionResult,
+  SlackIngressEnvelope,
+} from "../contracts.js";
+
+export type SlackIngressMode = "events_api" | "socket_mode";
+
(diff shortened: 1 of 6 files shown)


jonathanhaaswriter:
Superseded by #1948, which composes this contract into the validated portable continuity slice. Closing the duplicate lane to keep the stack linear.

PR #1940, opened 2026-07-16
Remove background context from policy proof

jonathanhaaswriter:
## Summary
- keep the context-free policy proof API for callers that do not execute graph queries
- share graph fixture validation between context-free and graph-store proof paths
- reserve caller-provided context for graph-store execution
- remove the structural violation introduced on main

## Validation
- `go test ./internal/testauthor/policy ./tools/testauthor -count=1`
- `make check-structural`
- `make changed-check`
- `git diff --check origin/main...HEAD`

PR #1943, opened 2026-07-16
Keep compatibility policy proof context-free

jonathanhaaswriter:
## Summary

- keep the no-store compatibility proof path context-free
- require caller-owned context only when graph fixtures execute through an injected store
- share graph fixture validation between compatibility and execution paths
- remove the remaining forbidden background context from the testauthor CLI

This replaces #1940 after `main` advanced with an overlapping patch.

## Validation

- `go test ./internal/testauthor/policy ./tools/testauthor -count=1`
- `make check-structural-build`
- `$(go env GOPATH)/bin/cerebrolint -nobackgroundctx ./internal/testauthor/policy ./tools/testauthor`
- `make check-structural`
- `git diff --check`


jonathanhaaswriter:
Superseded by #1950. The current main branch now threads caller context through expanded static validation and isolated graph-store proof batteries in these same files. Reapplying this patch would remove that new coverage and regress the current proof path, so this lane is closed rather than conflict-resolved.

PR #1952, opened 2026-07-16
fix(policy): restore policy candidate lint

jonathanhaaswriter:
## Summary
- preserve invalid-request classification while wrapping underlying parse and validation errors
- close policy-candidate query rows through the checked defer pattern
- restore the two policy-candidate lint shards currently failing on main

## Validation
- focused policy candidate, bootstrap, and PostgreSQL tests
- targeted golangci-lint for the three affected packages: 0 issues
- structural checks
- staged and range leak checks


jonathanhaaswriter:
Closing because current main now includes the lint corrections together with the newer policy experiment behavior. Applying this older patch would remove that behavior, so it should not land independently.

PR #2315, opened 2026-08-12
Fix entity catalog hosted lint gates

jonathanhaaswriter:
## Summary

- allow the bounded private entity-relation validator to use the same nine-field contract as its public store operation
- remove obsolete MCP raw-Cypher row conversion helpers left behind by the typed entity-catalog cutover

## Validation

- `make lint-shard LINT_PACKAGES="./internal/..." LINT_SHARD_TOTAL="4" LINT_SHARD_INDEX="1"`
- `go test ./internal/bootstrap -run TestMCPAssets -count=1`
- `git diff --check`

### DIFF
--- a/crates/organizational-store/src/neo4j.rs
+++ b/crates/organizational-store/src/neo4j.rs
@@ -3061,6 +3061,7 @@ fn validate_catalog_request(
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
 fn validate_catalog_relation_request(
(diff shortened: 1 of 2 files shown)


jonathanhaaswriter:
Superseded by the narrower fixes merged in #2312 and #2314. Current main removes the retired MCP parsers and replaces the oversized validator signature with grouped cursor fields, so this compatibility `allow` should not land.

PR #2323, opened 2026-08-13
Cover Slack phase report boundaries

jonathanhaaswriter:
## Summary
- cover empty phase-report inputs and percentile samples
- cover every closed endpoint, phase, and outcome wire variant
- restore the focused Slack black-box coverage floor without lowering it

## Validation
- `rustfmt --edition 2024 --check tools/slack-agent-blackbox/src/phase_report.rs`
- `git diff --check`

The Mac Cargo wrapper blocked the focused test before execution because the managed cache was 38.4 GB below its safety threshold. Hosted graph-actions is the authoritative focused coverage receipt.

### DIFF
--- a/tools/slack-agent-blackbox/src/phase_report.rs
+++ b/tools/slack-agent-blackbox/src/phase_report.rs
@@ -271,6 +271,57 @@ mod tests {
         );
     }
 
+    #[test]
+    fn empty_samples_and_receipts_fail_closed() {
+        assert_eq!(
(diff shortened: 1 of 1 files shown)


jonathanhaaswriter:
Superseded by #2322, which covers the same complete endpoint/phase/outcome label surface and empty percentile boundary on the same exact base. Keeping one hosted-check and merge path for the candidate fix.

PR #2587, opened 2026-08-25
Repair vendor register CI contracts

jonathanhaaswriter:
## Scope
- split the 32-field Go vendor row into cohesive embedded identity, governance, evidence, and queue records while preserving the flat JSON contract
- enable the Rust `time` macros used by date parsing
- construct current non-default Rust domain types explicitly in tests

## Local validation
- `go test ./internal/ports ./internal/sourcehttp/organizationalgraph ./internal/bootstrap -count=1`
- `make proto-lint check-structural`
- `git diff --check`

## Remaining generated-code gate
The merged vendor-register protobuf changed without regenerating the checked-in Rust Buffa and ConnectRPC outputs. `make proto-generate` is currently stopped by the managed Cargo capacity gate before the codegen tools can be installed (30,346,426,982 bytes short). This draft deliberately does not claim codegen completion; regeneration is the next action when DDESK storage or the managed cache is available.


jonathanhaaswriter:
Superseded by the complete forward repair in #2586, merged as 01d03eac016793577827e9ab366ba60763c5fffc. The remaining diff is an alternative representation of the same fixes and would reintroduce stale generated contracts, so this draft will not be merged.

PR #2598, opened 2026-08-25
feat(sources): complete AI catalog Rust authority

jonathanhaaswriter:
[...]

PR #2600, opened 2026-08-25
Add Langfuse runtime catalog fixtures

jonathanhaaswriter:
## Summary

- add deterministic runtime fixtures for the Langfuse `api_key` family
- add deterministic runtime fixtures for the Langfuse `project_member` family
- satisfy the catalog contract for every declared Langfuse runtime family

## Validation

- `make catalog-check`
- `go test ./sources/langfuse -count=1`
- `git diff --check`

The fixtures were generated through the existing Langfuse source adapters from their provider-shaped access-family test responses. No credentials or secret values are included.

### DIFF
--- a/sources/langfuse/testdata/discover_api_key.json
+++ b/sources/langfuse/testdata/discover_api_key.json
@@ -1,0 +1,3 @@
+[
+  "urn:cerebro:writer:langfuse_api_key:key_123"
+]
--- a/sources/langfuse/testdata/discover_project_member.json
+++ b/sources/langfuse/testdata/discover_project_member.json
@@ -1,0 +1,3 @@
+[
+  "urn:cerebro:writer:langfuse_project_member:member_123"
+]
--- a/sources/langfuse/testdata/read_api_key.json
+++ b/sources/langfuse/testdata/read_api_key.json
@@ -1,0 +1,29 @@
+[
+  {
+    "id": "langfuse-writer-40fd2d487aef-api_key-key_123",
+    "tenant_id": "writer",
+    "source_id": "langfuse",
+    "kind": "langfuse.api_key",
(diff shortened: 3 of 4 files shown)


jonathanhaaswriter:
Current `main` catalog failure is reproduced and the four missing files are the correct narrow scope. One deterministic issue remains in this exact head: the checked-in event IDs were generated from an ephemeral `httptest` base URL, so regenerating the same scoped source inputs changes both IDs. I verified this by two consecutive focused runs. Commit `a35e23424eb399b29bdd31a1a63857ba6134fd7f` on preserved branch `codex/langfuse-runtime-fixtures-deterministic` adds `sourcefixture.StabilizeEvents`, derives all four fixtures from the existing `api_key` and `project_member` source tests, and makes the comparison repeatable. Local receipts on that commit: full `go test ./sources/langfuse -count=1`, `go test ./tools/catalogcheck -count=1`, `go test ./internal/connectorcatalog -count=1`, `make catalog-check`, and `git diff --check` all pass. Please incorporate that stabilization or an equivalent forward fix on this canonical branch. I will not mutate this owned lane.

jonathanhaaswriter:
Ownership update: earlier draft #2598 already contains these same four Langfuse fixture paths and also carries the metrics-query bound, pagination extraction, and Rust module-formatting repairs required by the #2597 qualification. Its maintenance and catalog-source-catalog checks pass on exact head `4bef5d127ef801fa31ba03793be4405f1f5043a4`; only four explicit portable-AI Clippy findings remain. #2600 is therefore overlapping scope rather than the sole catalog repair. I will not modify or close either owned branch; please preserve the coherent #2598 lane and avoid landing duplicate fixture identities.

PR #2702, opened 2026-08-26
refactor(sources): retire Box Go projection writer

jonathanhaaswriter:
## Summary

- remove the production Box Go projection implementation for all closed families
- fail closed if the compatibility projector is invoked for a Rust-authoritative Box event
- preserve provider fixtures

## Authority proof

- the compiled Rust source catalog marks all 5 Box families collection-authoritative and projection-authoritative (pinned in #2699)
- the provider-specific Go source loader is already retired (TestBuiltinRetiresCoveredProviderGoLoaders)
- compatibility projection attempts now return a typed Rust-authority error before emitting graph writes

## Validation

- gofmt -l internal/sourceprojection
- go build ./internal/sourceprojection
- go test ./internal/sourceprojection -run Box -count=1 -v
- go test ./internal/sourceprojection -count=1
- go test ./internal/sourceregistry -run TestBuiltinRetiresCoveredProviderGoLoaders -count=1

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2709, opened 2026-08-26
test: repair main after provider projection retirements

jonathanhaaswriter:
## Summary

Main is red after the merged projection-retirement wave (#2697, #2700–#2703 and the codex retirements): three cross-package consumers still replayed retired providers through the Go projector, and one lint violation landed. This PR fixes all currently-failing Go tests and the lint finding on main.

- `internal/graphrebuild` and `internal/sourceruntime` organizational-platform benchmark corpora projected `box.content_assets`, which now fails closed after #2702 — moved to the live `dropbox_business.content_assets` family, which has identical generic-asset projection semantics (same entity shape and `resource_id` attributes).
- `cmd/projection-parity-adapter` catalog-contract fact test did the same — moved to `dropbox_business`.
- `newOpenAIOracleService` (from #2697) took a `graph` parameter every caller passed as nil — flagged by `unparam` in lint-internal — dropped the parameter.

## Validation

- go test ./internal/sourceprojection ./internal/sourceruntime ./internal/graphrebuild ./cmd/projection-parity-adapter -count=1
- golangci-lint run ./internal/sourceprojection (0 issues)
- gofmt -l over all touched packages (clean)
- go test ./internal/... ./cmd/... had no other failures on main before this fix (full-suite run; the four fixed above were the complete failure list)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2710, opened 2026-08-26
test(bench): move Go projection corpus off retired Box path

jonathanhaaswriter:
## Summary

- rebuild the Go organizational-platform benchmark corpus on `dropbox_business.content_assets`
- fixes `TestOrganizationalPlatformRawBenchmarkCorpus` (internal/sourceruntime) and `TestOrganizationalPlatformBenchmarkCorpus` (internal/graphrebuild), red on `main` since #2702 retired the Box Go projection writer
- unblocks the race/test shards on every open PR (they all inherit this failure through the merge ref)

## Why dropbox_business

- `dropbox_business.content_assets` uses the same generic asset projection template (`projection.template: asset`) as Box did, so the measured workload shape is unchanged: one entity per record, zero links
- `dropbox_business` is not in `TestBuiltinRetiresCoveredProviderGoLoaders` or any retirement lane, so the Go compatibility path stays live for this corpus
- the Rust side of `make rust-organizational-platform-benchmark` keeps its `box_assets` corpus — Box *is* Rust-authoritative, which is exactly why the Go side had to move; the paired comparison remains same-shape (generic raw asset projection), and the Go sub-benchmark labels are now `projection/content_assets/...`

## Validation

- gofmt -l internal/sourceruntime internal/graphrebuild
- go vet ./internal/sourceruntime ./internal/graphrebuild
- go test ./internal/sourceruntime -run TestOrganizationalPlatformRawBenchmarkCorpus -count=1
- go test ./internal/graphrebuild -run TestOrganizationalPlatformBenchmarkCorpus -count=1

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/graphrebuild/organizational_platform_benchmark_test.go
+++ b/internal/graphrebuild/organizational_platform_benchmark_test.go
@@ -25,7 +25,7 @@ func BenchmarkOrganizationalPlatformGo(b *testing.B) {
 	for _, recordCount := range []int{100, 1_000, 5_000} {
 		events := organizationalBenchmarkAssetEvents(recordCount)
 		verifyOrganizationalBenchmarkProjection(b, events)
-		b.Run(fmt.Sprintf("projection/box_assets/records_%d", recordCount), func(b *testing.B) {
+		b.Run(fmt.Sprintf("projection/content_assets/records_%d", recordCount), func(b *testing.B) {
 			b.ReportAllocs()
 			b.ResetTimer()
 			for range b.N {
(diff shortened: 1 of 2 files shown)


jonathanhaaswriter:
Exact head `358beb669c3beb69c75da436bead6afa551d2fb0` remains red in `race-15` job `98083627953`: `TestProjectRecordEmitsCatalogContractFact` in `cmd/projection-parity-adapter` still exercises `box.content_assets` and fails closed with `box projection requires Rust authority`. The two benchmark corpus repairs here cover current-main test-3/test-11 and race-13, but the adapter fixture needs the independent third replacement before this repair can make current main green. Broader #2709 contains that adapter change but is currently conflicting against `624891856f892bebebea35c374e86afddceba787`.

PR #2726, opened 2026-08-26
Fix Rust graph qualification route ownership

jonathanhaaswriter:
## Summary
- send graph qualification reads to the Rust product listener with tenant-bound authentication
- keep Go readiness checks and assert the retired Go graph route remains absent
- prove restart stability and fail-closed behavior against the runtime that owns each surface
- add focused shell authentication and route-ownership contracts

## Local validation
- `shellcheck scripts/qualify-rust-graph.sh scripts/lib/rust-tenant-auth.sh`
- `bash -n scripts/qualify-rust-graph.sh scripts/lib/rust-tenant-auth.sh`
- exact tenant-auth compatibility vector and short-secret rejection
- `rustfmt +1.93.1 --edition 2024 --check crates/cerebro-platform/tests/rust_only_runtime_contract.rs`
- `git diff HEAD^ --check`

Local Cargo execution was capacity-blocked before compilation; exact-head hosted Rust and Ephemeral Cerebro checks are authoritative.

PR #2732, opened 2026-08-26
test: align Cloudflare authority contract

jonathanhaaswriter:
## Summary

- align the checked-in Cloudflare runtime test with the catalog's landed authoritative collection state
- require every compiled Cloudflare family to remain collection-authoritative and projection-authoritative

The preceding Cloudflare catalog transition removed the invalid configurable `tenant_id` binding because tenant identity is supplied by the authenticated runtime. The source catalog and Go contract already require all Cloudflare families to be Rust-authoritative; this updates the remaining stale Rust assertion without changing production authority.

## Validation

- `go test ./internal/connectorcatalog -run TestBuiltinCloudflareCatalogClosesTheProviderFamilyAndCredentialContracts -count=1`
- `make catalog-check`
- `git diff --check`

Local Rust execution is blocked before start by the managed Cargo capacity guard: the DDESK checkout has no registered disk and the shared cache is about 34.3 GB short. Hosted exact-head Rust checks are required before merge.

### DIFF
--- a/crates/source-runtime-next/src/cloudflare/tests.rs
+++ b/crates/source-runtime-next/src/cloudflare/tests.rs
@@ -153,7 +153,7 @@ fn checked_in_catalog_compiles_the_closed_cloudflare_runtime() {
     .unwrap();
     let source = catalog.get("cloudflare").unwrap();
     assert_eq!(source.auth(), &AuthModel::BearerToken);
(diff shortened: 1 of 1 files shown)


jonathanhaaswriter:
The exact Cloudflare authority assertion change is already on `main` via #2726 (`aba27808f`). Closing this duplicate lane; no additional production behavior is being dropped.

PR #2735, opened 2026-08-26
Align Cloudflare authority regression test

jonathanhaaswriter:
## Summary

- Align the Cloudflare catalog regression test with the authoritative collection state produced by the checked-in catalog.
- Preserve the per-family projection-authority assertions.

## Test plan

- `rustup run 1.93.1 rustfmt --check crates/source-runtime-next/src/cloudflare/tests.rs`
- Hosted CI validates the Rust catalog and graph-action suites.


jonathanhaaswriter:
Closing this duplicate because #2732 already owns the same focused authority-test correction.

PR #2767, opened 2026-08-26
refactor(sources): retire Asana users Go projection

jonathanhaaswriter:
## Summary

- fail closed when the Go projector receives Asana user events
- preserve the existing Asana project and audit-event projection paths
- add a regression test proving the retired family cannot produce Go graph writes

## Validation

- go test ./internal/sourceprojection -count=1
- go test ./internal/sourceruntime ./internal/sourceruntime/sourceworker -count=1
- make check-structural check-arch
- make source-fixture-check
- make catalog-check

Rust formatting and test execution remain delegated to hosted checks because the managed local Cargo cache rejected the build before execution on its safe-capacity guard.

### DIFF
--- a/internal/sourceprojection/asana.go
+++ b/internal/sourceprojection/asana.go
@@ -1,14 +1,17 @@
 package sourceprojection
 
 import (
+	"errors"
 	"strings"
 
 	cerebrov1 "github.com/writer/cerebro/gen/cerebro/v1"
 	"github.com/writer/cerebro/internal/ports"
 )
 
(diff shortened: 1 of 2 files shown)


jonathanhaaswriter:
Post-merge qualification boundary: this change makes `asana.users` fail closed in the Go projector, but main does not yet contain evidence of an effective persisted Rust projection authority record for that tenant/source/family. Compile-time catalog authority, metadata-only Go registration, fixtures, and green hosted checks do not establish collection -> append -> Rust projection -> checkpoint -> product-read execution. The forward repair is the evidence-bound Rust authority-readiness seam now in progress; no rollback or Go fallback is authorized. Until that repair is merged and qualified, treat `asana.users` as an unproven runtime cutover and do not represent it as live-provider or live-user-path verified.

PR #2867, opened 2026-08-28
Fix automatic release proof and receipt validation

jonathanhaaswriter:
## Summary
- dispatch Rust Graph Replacement when an exact-head automatic release has no existing proof run
- fail closed without dispatch when proof enumeration cannot be completed
- verify main still matches the candidate before dispatch
- isolate the smoke-receipt negative test from an inherited runtime digest
- cover one-time dispatch, uncertain proof enumeration, stale-main behavior, and inherited release-test environments

## Validation
- Ruff formatting and linting for the workflow regression test
- Actionlint for the automatic stable-release workflow
- Zizmor pedantic validation with no findings
- Automatic stable-release workflow unit tests
- Smoke-receipt verifier with an inherited valid runtime digest
- Complete release-smoke suite, including product-release tests
- Git diff whitespace validation


jonathanhaaswriter:
Forward validation update:

- Automatic stable qualification admitted an exact-main candidate and passed its prerequisite proofs and signed ephemeral smoke.
- Stable publication then exposed an inherited-environment test isolation defect before publication. Candidate and signature verification passed; no tag or release was published.
- The current head clears the inherited runtime digest in the negative test that asserts the required-value failure.
- The inherited-environment regression and complete release-smoke suite pass locally.
- No mirror, deployment, or canary action was dispatched by this change.

PR #2976, opened 2026-09-02
feat(sources): close AbuseIPDB collection in Rust

jonathanhaaswriter:
## Summary

AbuseIPDB is catalog-authoritative with a complete credential-free Rust kernel (request planning, decoding, normalization, projection), but it had no bridge into the closed source-execution protocol, so durable collection still ran through the Go catalog runtime.

- Add `crates/source-runtime-next/src/abuseipdb/source_execution.rs`: adapters for the `reports` and `ip_addresses` families, modeled on the Abnormal Security bridge. The bridge reads only the public filters that belong to the selected family (`ip_address`, `max_age_in_days` for reports; `confidence_minimum`, `ip_version` for ip_addresses), because the host forwards every declared selector.
- Register the adapters in the shared dispatcher and add the `abuseipdb.key` credential operation: the trusted host applies the provider's `Key` header (`credentialHeader` in `host.go`) and the Rust contract allowlists the operation.
- Go: bind both families in `RustAuthoritativeFamily` (default `reports`) and `PreviewRustFamily`; forward the four declared public filters in `PublicExecutionConfigForSource`; select the `api_key` credential alias in `CredentialBinding`; retire the static Go loader.
- Ledger: `rust_fail_closed` kernel rules for durable pull and preview with refreshed selector digests.
- Also carries the small follow-up to #2973 that drops the `allow(dead_code)` / `allow(unused_imports)` attributes on the seven now-registered modules.

## Validation

- `cargo fmt --all`; `cargo test -p cerebro-source-runtime-next --locked` (abuseipdb and source_execution suites); `cargo clippy -p cerebro-source-runtime-next --all-targets --all-features --locked -- -D warnings`
- `go test ./internal/sourceruntime/sourceworker/... ./internal/sourceregistry/... ./tools/rustcarve/... -count=1`; `go vet ./internal/sourceruntime/sourceworker/...`
- `go run ./tools/sourceoperationauthoritycheck` (clean)
- Five new Rust tests: both-family planning, fixture decode and seal, pagination and typed provider failures, filter/cursor/duplicate/secret rejection, dispatcher registration. Two new Go tests: `Key` header application and AbuseIPDB public config filtering, plus selector and credential-alias table rows.


jonathanhaaswriter:
Superseded by merged PR #2982. A three-way merge with current main produces the exact current main tree, so this PR has no remaining unique change to integrate.

PR #2977, opened 2026-09-02
feat(sources): add ActiveCampaign source execution bridge

jonathanhaaswriter:
## Summary

- Add `crates/source-runtime-next/src/activecampaign/source_execution.rs`: a per-family `ActiveCampaignSourceExecutionAdapter` set (`ACTIVECAMPAIGN_SOURCE_EXECUTION_ADAPTERS`) that bridges the existing credential-free ActiveCampaign kernel into the closed source-execution protocol.
- The compiled plan mirrors the catalog contract for all five families (`accounts`, `automations`, `campaigns`, `contacts`, `users`): `GET /api/3/<family>`, record selector `$.<family>[*]`, `id_field` `id`, offset pagination at page size 100, and the family event kind / schema ref from `ActiveCampaignRuntimeDefinition`.
- Public configuration is limited to the required per-account `base_url` (validated by the kernel to an `https://<account>.api-us1.com` origin, which also becomes the allowed origin) and the optional `family` selector; a missing or invalid `base_url` fails closed as `MissingConfiguration`.
- Requests are planned under a new `activecampaign.api_token` credential operation because the provider authenticates with a bare `Api-Token` header (no `Bearer` scheme, not `X-Api-Key`). The bridge asserts the kernel's header/scheme at plan time and never carries credential bytes.
- `map_error` covers every `ActiveCampaignError` variant onto the shared `SourceExecutionError` taxonomy; `normalize::event_id` becomes `pub(super)` so identity validation reuses the kernel's tenant-, origin-, and family-scoped event id.
- Deliberately not registered: the adapter set is not wired into the shared dispatcher, `contract.rs` does not yet admit the credential operation, and the authority ledger is untouched. A follow-up PR wires all pending bridges at once, so the module is exposed with `#[allow(dead_code)]` / `#[allow(unused_imports)]` in the same shape earlier bridges used before registration.

## Validation

- `cargo fmt --all`
- `cargo test -p cerebro-source-runtime-next --locked activecampaign` (13 passed: 8 kernel tests plus 5 new bridge tests covering planning for every family without credentials, fixture decoding with the exact contract through `seal_page_program_v2`, offset pagination and typed provider failures including 429 with `retry-after`, rejection of bad cursors / wrong family / missing or invalid `base_url` / conflicting duplicates / credential-shaped nested keys / nested `tenant_id`, and adapter-set coverage of `ActiveCampaignFamily::ALL`)
- `cargo clippy -p cerebro-source-runtime-next --all-targets --all-features --locked -- -D warnings`


jonathanhaaswriter:
Superseded by merged PR #2982. A three-way merge with current main produces the exact current main tree, which also contains the expanded registered-adapter coverage.

PR #2978, opened 2026-09-02
feat(sources): add Acunetix source execution bridge

jonathanhaaswriter:
## Summary

- Add `crates/source-runtime-next/src/acunetix/source_execution.rs`: a self-contained bridge from the existing credential-free Acunetix kernel into the closed source-execution protocol, covering all five cataloged families (`reports`, `scanning_profiles`, `scans`, `targets`, `vulnerabilities`).
- The bridge reads only `base_url` (required; each tenant supplies its own HTTPS `/api/v1` origin, so the compiled plan carries a `https://{host}/api/v1` placeholder like the Aha bridge) and the optional `family` guard from public config. It plans origin-restricted `GET` requests with the catalog's `l`/`c` cursor pagination, decodes bounded pages through the kernel, and maps every `AcunetixError` variant onto the shared taxonomy.
- Acunetix authenticates with a bare `X-Auth` header (no scheme), so the bridge declares `CREDENTIAL_OPERATION = "acunetix.x_auth"` rather than reusing `source.bearer` / `source.x_api_key`.
- Expose `normalize::event_id` as `pub(super)` so the bridge can re-derive record identity.
- **Not registered yet.** The adapter set is gated with `allow(dead_code)` and the credential operation is not added to `contract.rs`; a follow-up PR wires every pending bridge into the shared dispatcher at once. No changes under `internal/`, the dispatcher, the contract, or the authority ledger.

## Validation

From the worktree root with `CARGO_TARGET_DIR` pointed at the shared target dir:

- `cargo fmt --all`
- `cargo test -p cerebro-source-runtime-next --locked acunetix` — 12 passed (7 kernel tests + 5 new bridge tests: planning every family without credentials, decoding every Go oracle fixture with the exact contract and sealing via `seal_page_program_v2`, pagination and typed provider failures including 429 with `retry-after`, rejection of a bad cursor / wrong `family` / missing or non-HTTPS `base_url`, duplicate collapse and conflict, nested credential keys and nested `tenant_id`, and adapter-set closure against `AcunetixFamily::ALL`)
- `cargo clippy -p cerebro-source-runtime-next --all-targets --all-features --locked -- -D warnings` — clean


jonathanhaaswriter:
Superseded by merged PR #2982. A three-way merge with current main produces the exact current main tree, which also contains the expanded registered-adapter coverage.

